PR #6, opened 2020-01-02
merge eloquent branch into master

adamsj-ros:
Still very much a WIP but added azure-pipelines.yml to use CI to continue development.


iguessthislldo:
I can't comment on it directly because it's not near something being changed, but these lines:

https://github.com/oci-labs/rmw_opendds/blob/920e07a437224751f357824ab1146b42073a62c8/rmw_opendds_cpp/CMakeLists.txt#L77-L78

should be replace by one or more `target_link_libraries` on the appropriate OpenDDS libraries that need to be used. There is an example of this in [this CMakelists file](https://github.com/objectcomputing/OpenDDS/blob/master/DevGuideExamples/DCPS/Messenger/CMakeLists.txt).

adamsj-ros:
> I can't comment on it directly because it's not near something being changed, but these lines:
> 
> https://github.com/oci-labs/rmw_opendds/blob/920e07a437224751f357824ab1146b42073a62c8/rmw_opendds_cpp/CMakeLists.txt#L77-L78
> 
> should be replace by one or more `target_link_libraries` on the appropriate OpenDDS libraries that need to be used. There is an example of this in [this CMakelists file](https://github.com/objectcomputing/OpenDDS/blob/master/DevGuideExamples/DCPS/Messenger/CMakeLists.txt).

With this PR, I'm trying to add consistency with the sibling repo, rosidl_typesupport_opendds. Since  this project currently fails to build and this cmake change would best be verified with a clean build, I'll add a new issue and move this PR forward.

rmw_opendds_cpp/src/rmw_take.cpp
```diff
@@ -24,8 +24,8 @@
 #include "rmw_opendds_cpp/identifier.hpp"
 
 // include patched generated code from the build folder
-#include "./opendds_static_serialized_dataSupport.h"
-#include "./opendds_static_serialized_data.h"
+#include "opendds_static_serialized_dataTypeSupportC.h"
+// #include "./opendds_static_serialized_data.h"
```
neeleym: This will eventually need to be removed. Unsure as to why it is still there.

azure-pipelines.yml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+- job: Linux
+  timeoutInMinutes: 90
+  pool:
+    vmImage: 'ubuntu-18.04'
+  container: adamsjoci/opendds_ros2_docker:latest
+  steps:
+  - script: |
+      printenv
+        echo hello
```
neeleym: Is the 'echo hello' really needed here?

neeleym (approved):

azure-pipelines.yml
adamsj-ros: The entire build step isn't really needed but I'm using it for now for trouble shooting. Will mark for removal.

rmw_opendds_cpp/src/rmw_take.cpp
adamsj-ros: removed in latest

PR #11, opened 2020-01-10
DDS namespace updates

adamsj-ros:
- differences found versus current version of this repos's source and other build errors
- other likely candidates (i.e. DDS_)

### DIFF
--- a/rmw_opendds_cpp/src/rmw_publisher.cpp
+++ b/rmw_opendds_cpp/src/rmw_publisher.cpp
@@ -96,7 +96,7 @@ rmw_create_publisher(
     RMW_SET_ERROR_MSG("node info handle is null");
     return NULL;
   }
-  auto participant = static_cast<DDSDomainParticipant *>(node_info->participant);
+  auto participant = static_cast<DDS::DomainParticipant *>(node_info->participant);
   if (!participant) {
     RMW_SET_ERROR_MSG("participant handle is null");
     return NULL;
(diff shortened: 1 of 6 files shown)


adamsj-ros:
@mitza-oci  and @neeleym, I found these DDS namespace changes from the rmw_connext. I think most of these are correct with the possible exception of DDS::Long and DDS::Octet.

neeleym (changes requested):
DDS::Long should be CORBA::Long
DDS::Octet should be CORBA::Octet
DDS::String_free should be CORBA::string_free

adamsj-ros:
> DDS::Octet should be CORBA::Octet

Is this true for IDL files too?

rmw_opendds_cpp/src/rmw_publisher.cpp
```diff
@@ -195,20 +195,20 @@ rmw_create_publisher(
 
     topic = participant->create_topic(
       topic_str, type_name.c_str(),
-      default_topic_qos, NULL, DDS_STATUS_MASK_NONE);
+      default_topic_qos, NULL, DDS::STATUS_MASK_NONE);
```
jwillemsen, line 198: See https://github.com/objectcomputing/OpenDDS/issues/1306, this mask is not yet there in OpenDDS

adamsj-ros: Good find. At this point, I'm only trying to make some changes that are mostly cosmetic. @neeleym is helping me review the real OpenDDS interface. I see that you're correct. We'll add this to our list to fix.

PR #13, opened 2020-01-27
WIP updates to node,publish, and publisher.

neeleym:
Simply committing some work in progress. Node, publish, and publisher were worked on here, but this will not build or run yet. A LOT of
vendor-specific leftovers in here.

### DIFF
--- a/rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/node.hpp
+++ b/rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/node.hpp
@@ -28,6 +28,12 @@ create_node(
   size_t domain_id,
   const rmw_node_security_options_t * options);
 
+
+RMW_OPENDDS_SHARED_CPP_PUBLIC
+        rmw_ret_t
+assert_liveliness(const char * implementation_identifier, const rmw_node_t * node);
+
+
 RMW_OPENDDS_SHARED_CPP_PUBLIC
 rmw_ret_t
 destroy_node(const char * implementation_identifier, rmw_node_t * node);
--- a/rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/node.hpp
+++ b/rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/node.hpp
@@ -30,7 +30,7 @@ create_node(
 
 
 RMW_OPENDDS_SHARED_CPP_PUBLIC
-        rmw_ret_t
+rmw_ret_t
 assert_liveliness(const char * implementation_identifier, const rmw_node_t * node);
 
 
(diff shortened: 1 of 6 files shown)


rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/node.hpp
```diff
@@ -28,6 +28,12 @@ create_node(
   size_t domain_id,
   const rmw_node_security_options_t * options);
 
+
+RMW_OPENDDS_SHARED_CPP_PUBLIC
+        rmw_ret_t
```
iguessthislldo:
```suggestion
rmw_ret_t
```

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/node_info_and_types.hpp
```diff
@@ -33,6 +33,16 @@ get_publisher_names_and_types_by_node(
   bool no_demangle,
   rmw_names_and_types_t * topic_names_and_types);
 
+RMW_OPENDDS_SHARED_CPP_PUBLIC
+        rmw_ret_t
```
iguessthislldo:
```suggestion
rmw_ret_t
```

iguessthislldo:
Also It looks like you guys are using GNU/ACE style. Is this the style of ROS or is this a left over from Johnny working on it? If ROS doesn't have a recommended style I think using the OpenDDS style would be a good idea.

adamsj-ros:
> Also It looks like you guys are using GNU/ACE style. Is this the style of ROS or is this a left over from Johnny working on it? If ROS doesn't have a recommended style I think using the OpenDDS style would be a good idea.

ROS offers uncrustify configs and linters to check and format the code. I feel like we can leave it as for now and run the scripts to beautify it later. To answer your question, it seems that much of ROS2 prefers this style as it matches the other RMWs.

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/node.hpp
adamsj-ros: Style updates are helpful. We can commit but we will run a beautify and linter at some point later in the development.

iguessthislldo: This isn't a style thing, this is a typo.

iguessthislldo: This is just a typo.

adamsj-ros: Missed that. Thanks.

adamsj-ros:
@neeleym I'm just getting around to checking this PR. I'm wondering which parts are still useful. I think our current plan is to comment out and leave much of the old code in for reference for those working on it now. Since you've removed large sections and replaced with TODOs, that may not be as helpful for this new effort. Any thoughts?

PR #14, opened 2020-02-20
Commented out code that prevents successful build

kuznetsovmoci:


adamsj-ros:
goal was a clean build and it appears that's what we have. Merging.

PR #15, opened 2020-02-24
pulled other missing runtime dependencies from rmw_connext

adamsj-ros:
I used rmw_connext as a template to remove the missing runtime dependencies. This is to help the test scripts execute properly. It's a stop gap until the actual functions can be checked against OpenDDS implementation.


kuznetsovmoci (commented):
Need some return statement (return RMW_RET_OK;) to rmw_node_assert_liveliness function

PR #16, opened 2020-02-25
initialization and node

lij-oci:


rmw_opendds_cpp/src/rmw_init.cpp
```diff
@@ -111,4 +121,5 @@ rmw_context_fini(rmw_context_t * context)
   *context = rmw_get_zero_initialized_context();
```
iguessthislldo: These lines are getting moved from `rmw_context_fini` to `rmw_shutdown`, so now except for the `RMW_CHECK_ARGUMENT_FOR_NULL` macro it looks like they're doing the same thing. Is that intended?

adamsj-ros:
This PR should end with a working `rmw_init` and `rmw_create_node` implemented with OpenDDS.

rmw_opendds_cpp/src/rmw_init.cpp
mitza-oci: @iguessthislldo was "indented" intended?

iguessthislldo: Yes... I probably misspelled it and went with whatever Chrome suggested.

rmw_opendds_shared_cpp/src/node.cpp
```diff
@@ -310,6 +311,9 @@ create_node(
   } catch (...) {}
 
   clean_node(dpf, node);
+  if (buf) {
+    rmw_free(buf);
+  }
```
iguessthislldo, line 316: Standard `free` will take a null pointer, so I doubt checking will be needed.

mitza-oci: http://docs.ros2.org/eloquent/api/rcutils/allocator_8h.html

iguessthislldo: > The default allocator uses malloc(), free(), calloc(), and realloc().

Okay, so yeah, it should be fine with a null pointer unless something else says otherwise.

adamsj-ros:
At this point the rmw_init and rmw_create_node seem to complete successfully. Once we have enough for a publisher, we can go back and verify the execution completely to publish data.

PR #19, opened 2020-03-13
required changes from rosidl_typesupport_opendds updates

adamsj-ros:
`rosidl_typesupport_opendds_cpp` is being updated ([PR #6](https://github.com/oci-labs/rosidl_typesupport_opendds/pull/6)) to more closely match to begin implementation of the new design.  This change is required for building `rmw_opendds` against the related change in `rosidl_typesupport_opendds_cpp`.

### DIFF
--- a/rmw_opendds_cpp/src/type_support_common.hpp
+++ b/rmw_opendds_cpp/src/type_support_common.hpp
@@ -90,7 +90,7 @@ _create_type_name(
   const std::string & sep)
 {
   return
-    std::string(callbacks->message_name) +
+    std::string(callbacks->message_namespace) +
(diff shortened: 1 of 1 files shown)


kuznetsovmoci (commented):
Message name is already included in the return string (one line below). Are you sure we need to replace package name with the message name here? If so - do we need message name in two places, or can simply remove the package name?

adamsj-ros:
> Message name is already included in the return string (one line below). Are you sure we need to replace package name with the message name here? If so - do we need message name in two places, or can simply remove the package name?

It was actually changed to message_namespace so I changed the rmw to match. However, rmw_connext has an entirely different block. I'll leave it this way for now until we get into the heart of this code.

PR #20, opened 2020-03-18
initial publish implementation

lij-oci:
Since OpenDDSStaticSerializedDataDataWriter is not a DataWriterImpl, casting to it will fail and writer->get_publication_id() will crash. A fake id is used here instead for the temporary testing purpose. Removed goto statements.

### DIFF
--- a/rmw_opendds_cpp/src/process_topic_and_service_names.cpp
+++ b/rmw_opendds_cpp/src/process_topic_and_service_names.cpp
@@ -22,6 +22,7 @@
 
 #include "rmw_opendds_shared_cpp/namespace_prefix.hpp"
 #include "rmw_opendds_shared_cpp/opendds_include.hpp"
+#include <iostream> //?? temp
 
 bool
 _process_topic_name(
@@ -29,6 +30,7 @@ _process_topic_name(
   bool avoid_ros_namespace_conventions,
   char ** topic_str)
 {
+  std::cout << "topic_name[" << topic_name << "], avoid_ros_namespace_conventions:" << avoid_ros_namespace_conventions << '\n'; //??
   bool success = true;
   rcutils_allocator_t allocator = rcutils_get_default_allocator();
   const char * topic_prefix = avoid_ros_namespace_conventions ? "" : ros_topic_prefix;
@@ -40,6 +42,7 @@ _process_topic_name(
     RMW_SET_ERROR_MSG("could not allocate memory for topic string");
     success = false;
   }
+  std::cout << "topic_prefix[" << topic_prefix << "], topic_str[" << *topic_str << "]\n"; //??
   return success;
 }
 
@@ -51,44 +54,27 @@ _process_service_name(
   char ** response_topic_str)
 {
   bool success = true;
+  std::cout << "service_name[" << service_name << "], avoid_ros_namespace_conventions:" << avoid_ros_namespace_conventions << '\n'; //??
   rcutils_allocator_t allocator = rcutils_get_default_allocator();
-
-  const char * requester_prefix = "";
-  const char * response_prefix = "";
-  char * request_concat_str = nullptr;
-  char * response_concat_str = nullptr;
-
-  if (!avoid_ros_namespace_conventions) {
-    requester_prefix = ros_service_requester_prefix;
-    response_prefix = ros_service_response_prefix;
-  }
+  const char * rq_pfx = avoid_ros_namespace_conventions ? "" : ros_service_requester_prefix;
+  const char * rp_pfx = avoid_ros_namespace_conventions ? "" : ros_service_response_prefix;
 
   // concat the ros_service_*_prefix and Request/Reply suffixes with the service_name
-  request_concat_str = rcutils_format_string(
-    allocator,
-    "%s%s%s", requester_prefix, service_name, "Request");
-  if (!request_concat_str) {
+  char * rq_str = rcutils_format_string(allocator, "%s%s%s", rq_pfx, service_name, "Request");
+  if (rq_str) {
+    *request_topic_str = CORBA::string_dup(rq_str);
+    allocator.deallocate(rq_str, allocator.state);
+  } else {
     RMW_SET_ERROR_MSG("could not allocate memory for request topic string");
     success = false;
-    goto end;
   }
-  response_concat_str = rcutils_format_string(
-    allocator,
-    "%s%s%s", response_prefix, service_name, "Reply");
-  if (!response_concat_str) {
+  char * rp_str = rcutils_format_string(allocator, "%s%s%s", rp_pfx, service_name, "Reply");
+  if (rp_str) {
+    *response_topic_str = CORBA::string_dup(rp_str);
+    allocator.deallocate(rp_str, allocator.state);
+  } else {
     RMW_SET_ERROR_MSG("could not allocate memory for response topic string");
     success = false;
-    goto end;
-  }
-  *request_topic_str = CORBA::string_dup(request_concat_str);
-  *response_topic_str = CORBA::string_dup(response_concat_str);
-
-end:
-  if (request_concat_str) {
-    allocator.deallocate(request_concat_str, allocator.state);
-  }
-  if (response_concat_str) {
-    allocator.deallocate(response_concat_str, allocator.state);
   }
   return success;
 }
--- a/rmw_opendds_cpp/src/process_topic_and_service_names.cpp
+++ b/rmw_opendds_cpp/src/process_topic_and_service_names.cpp
@@ -22,7 +22,6 @@
 
 #include "rmw_opendds_shared_cpp/namespace_prefix.hpp"
 #include "rmw_opendds_shared_cpp/opendds_include.hpp"
-#include <iostream> //?? temp
 
 bool
 _process_topic_name(
@@ -30,7 +29,6 @@ _process_topic_name(
   bool avoid_ros_namespace_conventions,
   char ** topic_str)
 {
-  std::cout << "topic_name[" << topic_name << "], avoid_ros_namespace_conventions:" << avoid_ros_namespace_conventions << '\n'; //??
   bool success = true;
   rcutils_allocator_t allocator = rcutils_get_default_allocator();
   const char * topic_prefix = avoid_ros_namespace_conventions ? "" : ros_topic_prefix;
@@ -42,7 +40,6 @@ _process_topic_name(
     RMW_SET_ERROR_MSG("could not allocate memory for topic string");
     success = false;
   }
-  std::cout << "topic_prefix[" << topic_prefix << "], topic_str[" << *topic_str << "]\n"; //??
   return success;
 }
 
@@ -54,7 +51,6 @@ _process_service_name(
   char ** response_topic_str)
 {
   bool success = true;
-  std::cout << "service_name[" << service_name << "], avoid_ros_namespace_conventions:" << avoid_ros_namespace_conventions << '\n'; //??
   rcutils_allocator_t allocator = rcutils_get_default_allocator();
   const char * rq_pfx = avoid_ros_namespace_conventions ? "" : ros_service_requester_prefix;
   const char * rp_pfx = avoid_ros_namespace_conventions ? "" : ros_service_response_prefix;
--- a/rmw_opendds_cpp/src/process_topic_and_service_names.cpp
+++ b/rmw_opendds_cpp/src/process_topic_and_service_names.cpp
@@ -31,25 +31,14 @@ _process_topic_name(
 {
   bool success = true;
   rcutils_allocator_t allocator = rcutils_get_default_allocator();
-
-  const char * topic_prefix = "";
-  char * concat_str = nullptr;
-
-  if (!avoid_ros_namespace_conventions) {
-    topic_prefix = ros_topic_prefix;
-  }
-
-  concat_str = rcutils_format_string(allocator, "%s%s", topic_prefix, topic_name);
-  if (!concat_str) {
+  const char * topic_prefix = avoid_ros_namespace_conventions ? "" : ros_topic_prefix;
+  char * concat_str = rcutils_format_string(allocator, "%s%s", topic_prefix, topic_name);
+  if (concat_str) {
+    *topic_str = CORBA::string_dup(concat_str);
+    allocator.deallocate(concat_str, allocator.state);
+  } else {
     RMW_SET_ERROR_MSG("could not allocate memory for topic string");
     success = false;
-    goto end;
-  }
-  *topic_str = CORBA::string_dup(concat_str);
-
-end:
-  if (concat_str) {
-    allocator.deallocate(concat_str, allocator.state);
   }
   return success;
 }
--- a/rmw_opendds_cpp/src/rmw_client.cpp
+++ b/rmw_opendds_cpp/src/rmw_client.cpp
@@ -50,7 +50,10 @@ rmw_create_client(
     node->implementation_identifier, opendds_identifier,
     return NULL)
 
-  RMW_OPENDDS_EXTRACT_SERVICE_TYPESUPPORT(type_supports, type_support, NULL)
+  const rosidl_service_type_support_t * type_support = rmw_get_service_type_support(type_supports);
+  if (!type_support) {
+    return NULL;
+  }
 
   if (!qos_profile) {
     RMW_SET_ERROR_MSG("qos_profile is null");
@@ -101,12 +104,12 @@ rmw_create_client(
     RMW_SET_ERROR_MSG("failed to allocate client");
     goto fail;
   }
-
+/*
   if (!get_datareader_qos(participant, *qos_profile, datareader_qos)) {
     // error string was set within the function
     goto fail;
   }
-/*
+
   //?? get_default_publisher_qos or pass DDS::Publisher* to get_datawriter_qos
   if (!get_datawriter_qos(participant, *qos_profile, datawriter_qos)) {
     // error string was set within the function
--- a/rmw_opendds_cpp/src/rmw_client.cpp
+++ b/rmw_opendds_cpp/src/rmw_client.cpp
@@ -106,12 +106,13 @@ rmw_create_client(
     // error string was set within the function
     goto fail;
   }
-
+/*
+  //?? get_default_publisher_qos or pass DDS::Publisher* to get_datawriter_qos
   if (!get_datawriter_qos(participant, *qos_profile, datawriter_qos)) {
     // error string was set within the function
     goto fail;
   }
-
+*/
   // allocating memory for request topic and response topic strings
   if (!_process_service_name(
       service_name,
--- a/rmw_opendds_cpp/src/rmw_publish.cpp
+++ b/rmw_opendds_cpp/src/rmw_publish.cpp
@@ -21,20 +21,24 @@
 #include "rmw_opendds_cpp/opendds_static_publisher_info.hpp"
 #include "rmw_opendds_cpp/identifier.hpp"
 
-//#include "dds/DCPS/DataWriterImpl.h"
-#include <dds/DdsDcpsGuidC.h>
-#include <dds/DCPS/RepoIdBuilder.h>
+#include "dds/DCPS/DataWriterImpl.h"
 #include "opendds_static_serialized_dataTypeSupportC.h"
 
 #include <ace/Message_Block.h>
 
-OpenDDS::DCPS::RepoId createID(long participantId, long key, CORBA::Octet kind) {
-  OpenDDS::DCPS::RepoIdBuilder idBd;
-  idBd.federationId(0x12345678);     // guidPrefix1
-  idBd.participantId(participantId); // guidPrefix2
-  idBd.entityKey(key);
-  idBd.entityKind(kind);
-  return idBd;
+bool set_key(OpenDDSStaticSerializedData & instance, DDS::DataWriter * dds_data_writer) {
+  OpenDDS::DCPS::DataWriterImpl* dwImpl = dynamic_cast<OpenDDS::DCPS::DataWriterImpl*>(dds_data_writer);
+  if (!dwImpl) {
+    RMW_SET_ERROR_MSG("failed to cast dds_data_writer to DataWriterImpl");
+    return false;
+  }
+  OpenDDS::DCPS::Message_Block_Ptr key(new ACE_Message_Block(KEY_HASH_LENGTH_16, ACE_Message_Block::MB_DATA, 0, 0, 0, 0));
+  OpenDDS::DCPS::Serializer ser(key.get());
+  ser << dwImpl->get_publication_id();
+  instance.serialized_key.length(KEY_HASH_LENGTH_16);
+  std::memcpy(instance.key_hash, key.get(), KEY_HASH_LENGTH_16);
+  std::memcpy(instance.serialized_key.get_buffer(), key.get(), KEY_HASH_LENGTH_16);
+  return true;
 }
 
 bool
@@ -52,14 +56,10 @@ publish(DDS::DataWriter * dds_data_writer, const rcutils_uint8_array_t * cdr_str
   }
 
   OpenDDSStaticSerializedData instance;
-  // TODO: This implementation is temporary until typesupport is finalized.
-  // Populate instance
-  // Since OpenDDSStaticSerializedDataDataWriter is not a DataWriterImpl, a fake id is used here for testing.
-  OpenDDS::DCPS::RepoId id = createID(0x11111111, 0x123456, OpenDDS::DCPS::ENTITYKIND_USER_WRITER_WITH_KEY);
-  instance.serialized_key.length(KEY_HASH_LENGTH_16);
+  if (!set_key(instance, dds_data_writer)) {
+    return false;
+  }
   instance.serialized_data.length(static_cast<CORBA::ULong>(cdr_stream->buffer_length));
-  std::memcpy(instance.key_hash, &id, KEY_HASH_LENGTH_16);
-  std::memcpy(instance.serialized_key.get_buffer(), &id, KEY_HASH_LENGTH_16);
   std::memcpy(instance.serialized_data.get_buffer(), cdr_stream->buffer, cdr_stream->buffer_length);
 
   DDS::ReturnCode_t status = writer->write(instance, DDS::HANDLE_NIL);
--- a/rmw_opendds_cpp/src/rmw_publish.cpp
+++ b/rmw_opendds_cpp/src/rmw_publish.cpp
@@ -60,6 +60,7 @@ publish(DDS::DataWriter * dds_data_writer, const rcutils_uint8_array_t * cdr_str
   instance.serialized_data.length(static_cast<CORBA::ULong>(cdr_stream->buffer_length));
   std::memcpy(instance.serialized_data.get_buffer(), cdr_stream->buffer, cdr_stream->buffer_length);
 
+  std::cout << "publish writer->write\n";
   DDS::ReturnCode_t status = writer->write(instance, DDS::HANDLE_NIL);
   return status == DDS::RETCODE_OK;
 }
@@ -72,6 +73,7 @@ rmw_publish(
   const void * ros_message,
   rmw_publisher_allocation_t * allocation)
 {
+  std::cout << "rmw_publish----->\n";
   if (!publisher) {
     RMW_SET_ERROR_MSG("publisher is null");
     return RMW_RET_ERROR;
@@ -102,8 +104,9 @@ rmw_publish(
 
   auto ret = RMW_RET_ERROR;
   rcutils_uint8_array_t cdr_stream = rcutils_get_zero_initialized_uint8_array();
-  cdr_stream.allocator = rcutils_get_default_allocator(); //?? not using the given allocation
+  cdr_stream.allocator = rcutils_get_default_allocator();
   try {
+/*
     if (!callbacks->to_cdr_stream(ros_message, &cdr_stream)) {
       throw std::string("failed to convert ros_message to cdr stream");
     }
@@ -113,6 +116,7 @@ rmw_publish(
     if (!cdr_stream.buffer) {
       throw std::string("no serialized message attached");
     }
+*/
     if (!publish(writer, &cdr_stream)) {
       throw std::string("failed to publish message");
     }
@@ -123,6 +127,7 @@ rmw_publish(
     RMW_SET_ERROR_MSG("rmw_publish failed");
   }
   cdr_stream.allocator.deallocate(cdr_stream.buffer, cdr_stream.allocator.state);
+  std::cout << "rmw_publish return " << ret << '\n'; //??
   return ret;
 }
 
--- a/rmw_opendds_cpp/src/rmw_publish.cpp
+++ b/rmw_opendds_cpp/src/rmw_publish.cpp
@@ -21,17 +21,27 @@
 #include "rmw_opendds_cpp/opendds_static_publisher_info.hpp"
 #include "rmw_opendds_cpp/identifier.hpp"
 
-#include "dds/DCPS/DataWriterImpl.h"
+//#include "dds/DCPS/DataWriterImpl.h"
+#include <dds/DdsDcpsGuidC.h>
+#include <dds/DCPS/RepoIdBuilder.h>
 #include "opendds_static_serialized_dataTypeSupportC.h"
 
 #include <ace/Message_Block.h>
 
+OpenDDS::DCPS::RepoId createID(long participantId, long key, CORBA::Octet kind) {
+  OpenDDS::DCPS::RepoIdBuilder idBd;
+  idBd.federationId(0x12345678);     // guidPrefix1
+  idBd.participantId(participantId); // guidPrefix2
+  idBd.entityKey(key);
+  idBd.entityKind(kind);
+  return idBd;
+}
+
 bool
 publish(DDS::DataWriter * dds_data_writer, const rcutils_uint8_array_t * cdr_stream)
 {
-  OpenDDSStaticSerializedDataDataWriter * data_writer =
-    OpenDDSStaticSerializedDataDataWriter::_narrow(dds_data_writer);
-  if (!data_writer) {
+  OpenDDSStaticSerializedDataDataWriter * writer = OpenDDSStaticSerializedDataDataWriter::_narrow(dds_data_writer);
+  if (!writer) {
     RMW_SET_ERROR_MSG("failed to narrow data writer");
     return false;
   }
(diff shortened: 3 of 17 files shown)


lij-oci:
Some thoughts:

1) Some hash code should be used to set key_hash and serialized_key instead of using publication guid, which is not required by RMW and which may have security issues.

2) So far, the DataWriterImpl is used only for get_publication_id(). Since it is an implementation detail, its use is factored out together with the key setting code in set_key().

rmw_opendds_cpp/src/rmw_publish.cpp
```diff
@@ -26,12 +26,26 @@
...
+  OpenDDS::DCPS::DataWriterImpl* dwImpl = dynamic_cast<OpenDDS::DCPS::DataWriterImpl*>(dds_data_writer);
+  if (!dwImpl) {
+    RMW_SET_ERROR_MSG("failed to cast dds_data_writer to DataWriterImpl");
+    return false;
+  }
+  OpenDDS::DCPS::Message_Block_Ptr key(new ACE_Message_Block(KEY_HASH_LENGTH_16, ACE_Message_Block::MB_DATA, 0, 0, 0, 0));
+  OpenDDS::DCPS::Serializer ser(key.get());
+  ser << dwImpl->get_publication_id();
+  instance.serialized_key.length(KEY_HASH_LENGTH_16);
+  std::memcpy(instance.key_hash, key.get(), KEY_HASH_LENGTH_16);
```
mitza-oci: This doesn't look right to me.  The type of the 2nd argument to memcpy is `ACE_Message_Block*`.  What was probably intended was the actual data that the `ACE_Message_Block` manages.  You can do that with `key->rd_ptr()`.
Another approach here would be to remove `ACE_Message_Block` and `Serializer` from this method.  Those are powerful tools that can turn any IDL-defined type into a byte buffer (or back the other way).  In this case all that power isn't really needed since type type returned by `get_publication_id` (a GUID) is a POD type.  It's just 16 contiguous bytes that you can directly access with `memcpy`.

rmw_opendds_cpp/src/rmw_publish.cpp
```diff
@@ -26,12 +26,26 @@
...
+  return true;
+}
+
 bool
 publish(DDS::DataWriter * dds_data_writer, const rcutils_uint8_array_t * cdr_stream)
 {
-  OpenDDSStaticSerializedDataDataWriter * data_writer =
-    OpenDDSStaticSerializedDataDataWriter::_narrow(dds_data_writer);
-  if (!data_writer) {
+  OpenDDSStaticSerializedDataDataWriter * writer = OpenDDSStaticSerializedDataDataWriter::_narrow(dds_data_writer);
```
mitza-oci: Use a `_var` smart pointer to avoid leaking memory.

rmw_opendds_cpp/src/rmw_publisher.cpp
```diff
@@ -110,195 +130,151 @@ rmw_create_publisher(
...
-      publisher->implementation_identifier = opendds_identifier;
-      publisher->data = publisher_info;
-      publisher->topic_name = reinterpret_cast<const char*>(rmw_allocate(strlen(topic_name) + 1));
-      if (!publisher->topic_name) {
-        throw std::string("failed to allocate memory for node name");
-      }
-      memcpy(const_cast<char*>(publisher->topic_name), topic_name, strlen(topic_name) + 1);
+    // update node_info
+    std::string mangled_name = qos_policies->avoid_ros_namespace_conventions ?
+      topic_name : publisher_info->topic_writer_->get_topic()->get_name();
```
mitza-oci: The object returned by `get_topic` is leaked here.  Use a  `_var` smart pointer.
The string returned by `get_name` should also be managed by a `_var` smart pointer.

rmw_opendds_cpp/src/type_support_common.hpp
```diff
@@ -55,6 +55,23 @@
...
+inline const rosidl_message_type_support_t *
+rmw_get_message_type_support(const rosidl_message_type_support_t * type_support)
+{
+  if (!type_support) {
+    RMW_SET_ERROR_MSG("type support is null");
+    return nullptr;
+  }
+  const rosidl_message_type_support_t * ts =
+    //get_message_typesupport_handle(type_support, rosidl_typesupport_opendds_cpp::typesupport_identifier);
+    get_message_typesupport_handle(type_support, "rosidl_typesupport_c"); //for temporary testing
```
mitza-oci: How do we determine what this string (2nd argument of get_message_typesupport_handle) should be?

mitza-oci (changes requested):

rmw_opendds_cpp/src/rmw_publish.cpp
```diff
@@ -26,12 +26,26 @@
...
+  if (!dwImpl) {
+    RMW_SET_ERROR_MSG("failed to cast dds_data_writer to DataWriterImpl");
+    return false;
+  }
+  OpenDDS::DCPS::Message_Block_Ptr key(new ACE_Message_Block(KEY_HASH_LENGTH_16, ACE_Message_Block::MB_DATA, 0, 0, 0, 0));
+  OpenDDS::DCPS::Serializer ser(key.get());
+  ser << dwImpl->get_publication_id();
+  instance.serialized_key.length(KEY_HASH_LENGTH_16);
+  std::memcpy(instance.key_hash, key.get(), KEY_HASH_LENGTH_16);
+  std::memcpy(instance.serialized_key.get_buffer(), key.get(), KEY_HASH_LENGTH_16);
```
sonndinh: Just a minor comment that swapping lines 39 and 40 might make the code easier to follow since lines 38 and 40 are both modifying `instance.serialized_key`.

sonndinh:
The header guards of the header files opendds_static_*.hpp in folder include/rmw_opendds_cpp/ still contain CONNEXT. We should replace them with OPENDDS. 

Good catch. I'll replace them.

rmw_opendds_cpp/src/rmw_publish.cpp
lij-oci: So if I replace the fake id with dwImpl->get_publication_id() in my previous implementation, do you mean something like the following?

  OpenDDS::DCPS::RepoId id = dwImpl->get_publication_id();
  std::memcpy(instance.key_hash, &id, KEY_HASH_LENGTH_16);
  instance.serialized_key.length(KEY_HASH_LENGTH_16);
  std::memcpy(instance.serialized_key.get_buffer(), &id, KEY_HASH_LENGTH_16);

mitza-oci: Yes, looks good from here

rmw_opendds_cpp/src/rmw_publish.cpp
lij-oci: Yeah, I see what you mean. The primary intention is to keep the order of the declaration: 
@topic struct OpenDDSStaticSerializedData {
  @key OctetArray16       key_hash;
  DDS::OctetSeq  serialized_key;
  DDS::OctetSeq  serialized_data;
};

To that end, the code is like this:
OpenDDS::DCPS::RepoId id = dwImpl->get_publication_id();
std::memcpy(instance.key_hash, &id, KEY_HASH_LENGTH_16);
instance.serialized_key.length(KEY_HASH_LENGTH_16);
std::memcpy(instance.serialized_key.get_buffer(), &id, KEY_HASH_LENGTH_16);

If I don't factor out this function, the above code would be immediately followed by these two lines to complete the initialization:
instance.serialized_data.length(static_cast<CORBA::ULong>(cdr_stream->buffer_length));
std::memcpy(instance.serialized_data.get_buffer(), cdr_stream->buffer, cdr_stream->buffer_length);

rmw_opendds_cpp/src/rmw_publisher.cpp
lij-oci: Do we have a smart pointer like _var for char* ? I searched dds spec, our dev guide, and our code, but didn't find it. It seems we have these options:

1. use std::unique_ptr, and provide a deleter to do delete[];
2. just use char*, and manually free it with CORBA::string_free().
3. define one, such as string_var. Two issues with this option: 1) where to define it, in OpenDDS or rmw? 2) is it worth the effort, compared to changing the functions to return std:string (the standard C++ way to replace C strings)?

Currently, I changed the code with option 2 as it's done in some other places.

mitza-oci: String_var is defined already by the specs and libraries that implement them.  Please use it.

lij-oci: Thanks.

rmw_opendds_cpp/src/rmw_publisher.cpp
```diff
@@ -110,195 +130,151 @@ rmw_create_publisher(
...
+    publisher->can_loan_messages = false;
 
-      publisher->can_loan_messages = false;
+    // create OpenDDSStaticPublisherInfo
+    buf = rmw_allocate(sizeof(OpenDDSStaticPublisherInfo));
+    if (!buf) {
+      throw std::string("failed to allocate memory for publisher info");
+    }
+    OpenDDSStaticPublisherInfo * publisher_info = nullptr;
+    RMW_TRY_PLACEMENT_NEW(publisher_info, buf, throw 1, OpenDDSStaticPublisherInfo,)
```
sonndinh, line 190: Should we throw a string specifying calling to this macro fails?

rmw_opendds_cpp/src/rmw_publisher.cpp
```diff
@@ -110,195 +130,151 @@ rmw_create_publisher(
...
-      // TODO: Register TypeSupport here ????????????????????????
-      if (ts->register_type(participant, "") != DDS::RETCODE_OK) {
-        throw std::string("failed to register OpenDDS type");
-      }
+    // create OpenDDSPublisherListener
+    buf = rmw_allocate(sizeof(OpenDDSPublisherListener));
+    if (!buf) {
+      throw std::string("failed to allocate memory for publisher listener");
+    }
+    RMW_TRY_PLACEMENT_NEW(publisher_info->listener_, buf, throw 1, OpenDDSPublisherListener,)
```
sonndinh, line 199: Similar comment to line 177.

rmw_opendds_cpp/src/rmw_publisher.cpp
```diff
@@ -110,195 +130,151 @@ rmw_create_publisher(
...
 
-      DDS::PublisherQos publisher_qos;
-      DDS::ReturnCode_t status = participant->get_default_publisher_qos(publisher_qos);
+    // create or find DDS::Topic
+    DDS::Topic_var topic;
+    if (!_process_topic_name(topic_name, qos_policies->avoid_ros_namespace_conventions, &topic_str)) {
+      throw std::string("failed to allocate memory for topic_str");
+    }
+    DDS::TopicDescription_var topic_description = participant->lookup_topicdescription(topic_str);
+    std::string type_name = _create_type_name(callbacks, "msg");
```
adamsj-ros: `_create_type_name` needs to be fixed. When I run the example (i.e. `./rmw_build/run_test.sh`), I get a seg fault inside `_create_type_name`

rmw_opendds_cpp/src/rmw_publisher.cpp
```diff
@@ -72,34 +97,29 @@ rmw_create_publisher(
...
   }
+
   auto participant = static_cast<DDS::DomainParticipant *>(node_info->participant);
   if (!participant) {
-    RMW_SET_ERROR_MSG("participant handle is null");
+    RMW_SET_ERROR_MSG("participant is null");
+    return NULL;
+  }
+
+//RMW_OPENDDS_EXTRACT_MESSAGE_TYPESUPPORT(type_supports, type_support, NULL)
```
adamsj-ros: Is this macro needed or are we waiting on type support? If we're waiting, we should involve @kuznetsovmoci and  @sonndinh to understand the problem in `rmw_create_publisher` at this point.

rmw_opendds_cpp/src/rmw_publisher.cpp
```diff
@@ -72,34 +97,29 @@ rmw_create_publisher(
...
+  auto type_support = rmw_get_message_type_support(type_supports);
+  if (!type_support) {
     return NULL;
   }
```
adamsj-ros: Working as expected? @kuznetsovmoci and @sonndinh?

rmw_opendds_cpp/src/rmw_publisher.cpp
```diff
@@ -110,195 +130,151 @@ rmw_create_publisher(
...
 
-  // TODO: TypeCode and susequent call to register external type have been removed
-  // May need to implement something similar after IDL type support is finalized
+  // TODO: Register TypeSupport here ????????????????????????
+  OpenDDSStaticSerializedDataTypeSupport_var ts = new OpenDDSStaticSerializedDataTypeSupportImpl();
```
adamsj-ros, line 133: working as expected?

rmw_opendds_cpp/src/rmw_publisher.cpp
```diff
@@ -110,195 +130,151 @@ rmw_create_publisher(
...
-    try {
-      // Begin initializing elements
-      publisher = rmw_publisher_allocate();
-      if (!publisher) {
-        throw std::string("failed to allocate publisher");
-      }
+    // create rmw_publisher_t
+    publisher = rmw_publisher_allocate();
+    if (!publisher) {
+      throw std::string("failed to allocate publisher");
```
adamsj-ros, line 170: While throw's aren't syntactically a goto, this code operates in much the same way. This isn't necessarily a critique. It's more of an observation.

adamsj-ros (commented):
Code looks and runs a lot better. We should involve @sonndinh and @kuznetsovmoci to figure out the type issues for publisher. At a minimum we should find a way to handle string serialization since it's the example code. That would allow us to verify the node and init functions too.

rmw_opendds_cpp/src/rmw_publisher.cpp
```diff
@@ -307,7 +283,7 @@ rmw_publisher_count_matched_subscriptions(
   size_t * subscription_count)
 {
   if (!publisher) {
-    RMW_SET_ERROR_MSG("publisher handle is null");
+    RMW_SET_ERROR_MSG("publisher is null");
```
sonndinh, line 269: Should we use macro `RMW_CHECK_ARGUMENT_FOR_NULL` to check for null arguments in all places for consistency?

sonndinh:
> The header guards of the header files opendds_static_*.hpp in folder include/rmw_opendds_cpp/ still contain CONNEXT. We should replace them with OPENDDS.
> 
> Good catch. I'll replace them.

There are a couple of places in rmw_publisher.cpp where "connext" is used in error messages.

rmw_opendds_cpp/src/rmw_publish.cpp
sonndinh: Yes, I agree.

rmw_opendds_cpp/src/rmw_publisher.cpp
mitza-oci: Yes, I thought about that issue when I reviewed it last week.  We may need to do another pass (after this PR) with a different error handling/cleanup approach.

rmw_opendds_cpp/src/rmw_publisher.cpp
mitza-oci: What's failing?

rmw_opendds_cpp/src/type_support_common.hpp
```diff
@@ -55,6 +55,23 @@
...
   }
 
+inline const rosidl_message_type_support_t *
+rmw_get_message_type_support(const rosidl_message_type_support_t * type_support)
+{
+  if (!type_support) {
+    RMW_SET_ERROR_MSG("type support is null");
+    return nullptr;
+  }
+  const rosidl_message_type_support_t * ts =
```
mitza-oci: Since the OpenDDS implementation of ROSIDL TypeSupport is not really ready for use, consider an interim implementation of this method (`rmw_get_message_type_support`) that returns a "dummy" version of `rosidl_message_type_support_t`

lij-oci:
As I move toward the end of the rmw_create_publisher function, 
there is a call to get_datawriter_qos() which is defined in 
opendds_rmw_ws/src/rmw_opendds/rmw_opendds_shared_cpp/src/qos.cpp:

bool
get_datawriter_qos(
  DDS::DomainParticipant * participant,
  const rmw_qos_profile_t & qos_profile,
  DDS::DataWriterQos & datawriter_qos)
{
  DDS::ReturnCode_t status; // @todo jwi = participant->get_default_datawriter_qos(datawriter_qos);
  if (status != DDS::RETCODE_OK) {
    RMW_SET_ERROR_MSG("failed to get default datawriter qos");
    return false;
  }

  if (!set_entity_qos_from_profile(qos_profile, datawriter_qos)) {
    return false;
  }

  return true;
}

I uncommented the code participant->get_default_datawriter_qos(datawriter_qos), and got:
error: ‘class DDS::DomainParticipant’ has no member named ‘get_default_datawriter_qos’;

I wonder whether we should pass DDS::Publisher* instead of DDS::DomainParticipant* into this function and call publisher->get_default_datawriter_qos(datawriter_qos).

rmw_opendds_cpp/src/type_support_common.hpp
lij-oci: In the original macro (see below), a rosidl_message_type_support_t is initialized by calling 
rmw_get_message_type_support first with rosidl_typesupport_opendds_c__identifier,
and then (if returned null) with rosidl_typesupport_opendds_cpp::typesupport_identifier.

These calls will end up in strcmp of these parameters with the passed-in 
rosidl_message_type_support_t->rosidl_message_typesupport_handle_function->typesupport_identifier.

The current value of that typesupport_identifier in my test runs is "rosidl_typesupport_c". 
It is not our opendds specific and could not match either our c or c++ typesupport_identifier.

At this point, I'm not sure how rmw framework passes that value here or 
whether it passes that value as default when no implementation specific value is provided, or
how an implementation should provide its typesupport_identifier, through configuration, file or code.

I took Jeremy's suggestion and made an inline function which made debugging easier.
I only did the cpp part for now and even that is commented out and a temp string is used to pass strcmp.
I'm going to run some other implementation code and hope to get some answers.


#define RMW_OPENDDS_EXTRACT_MESSAGE_TYPESUPPORT(TYPE_SUPPORTS, TYPE_SUPPORT, RET_VAL) \
  if (!TYPE_SUPPORTS) { \
    RMW_SET_ERROR_MSG("type supports handle is null"); \
    return RET_VAL; \
  } \
  const rosidl_message_type_support_t * TYPE_SUPPORT = \
    get_message_typesupport_handle( \
    TYPE_SUPPORTS, rosidl_typesupport_opendds_c__identifier); \
  if (!TYPE_SUPPORT) { \
    TYPE_SUPPORT = get_message_typesupport_handle( \
      TYPE_SUPPORTS, rosidl_typesupport_opendds_cpp::typesupport_identifier); \
    if (!TYPE_SUPPORT) { \
      char __msg[1024]; \
      snprintf( \
        __msg, 1024, \
        "type support handle implementation '%s' (%p) does not match valid type supports " \
        "('%s' (%p), '%s' (%p))", \
        TYPE_SUPPORTS->typesupport_identifier, \
        static_cast<const void *>(TYPE_SUPPORTS->typesupport_identifier), \
        rosidl_typesupport_opendds_cpp::typesupport_identifier, \
        static_cast<const void *>(rosidl_typesupport_opendds_cpp::typesupport_identifier), \
        rosidl_typesupport_opendds_c__identifier, \
        static_cast<const void *>(rosidl_typesupport_opendds_c__identifier)); \
      RMW_SET_ERROR_MSG(__msg); \
      return RET_VAL; \
    } \
  }

lij-oci: GitHub erased all the \ character in the macro.

rmw_opendds_cpp/src/rmw_publish.cpp
adamsj-ros: @mitza-oci: Is this resolved?

rmw_opendds_cpp/src/rmw_publish.cpp
adamsj-ros: @mitza-oci : Why in this case isn't it showing the updated code for this comment? I still see the raw pointer.

rmw_opendds_cpp/src/rmw_publisher.cpp
adamsj-ros: Resolved?

jwillemsen:
> I wonder whether we should pass DDS::Publisher* instead of DDS::DomainParticipant* into this function and call publisher->get_default_datawriter_qos(datawriter_qos).

Yes, see https://github.com/ros2/rmw_connext/issues/318

lij-oci:
The effort of replacing the crashing code with temporary dummy code or commenting out has been made through publisher, publish, service, client, subscription and some shared files. The MinimalPublisher test has run through all the above in Node constructor, run into MinimalPublisher constructor body, and passed create_publisher(); It terminated in create_wall_timer().
```
root@e1ff29601fe4:/opt/workspace# rmw_build/run_test.sh
GNU gdb (Ubuntu 8.1-0ubuntu3.2) 8.1.0.20180409-git
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /opt/workspace/install/examples_rclcpp_minimal_publisher/lib/examples_rclcpp_minimal_publisher/publisher_member_function...done.
Starting program: /opt/workspace/install/examples_rclcpp_minimal_publisher/lib/examples_rclcpp_minimal_publisher/publisher_member_function 
warning: Error disabling address space randomization: Operation not permitted
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7fdf41ae1700 (LWP 84732)]
[New Thread 0x7fdf412e0700 (LWP 84733)]
rclcpp::init() done
create_node(rmw_opendds_cpp, minimal_publisher, /, 0)
qos.user_data.value[name=minimal_publisher;namespace=/;]
[New Thread 0x7fdf3bfff700 (LWP 84734)]
[New Thread 0x7fdf3b7fe700 (LWP 84735)]
[New Thread 0x7fdf3a9bc700 (LWP 84736)]
[New Thread 0x7fdf3a1bb700 (LWP 84737)]
return node <-------
type_support(rosidl_typesupport_c, 0x7fdf4540d140, 1)
topic_name[/rosout], avoid_ros_namespace_conventions:0
topic_prefix[rt], topic_str[rt/rosout]
type_name[callbacks->message_namespace::msg::dds_::callbacks->message_name_], topic_name[/rosout], topic_str[rt/rosout]
[New Thread 0x7fdf399ba700 (LWP 84738)]
[New Thread 0x7fdf391b9700 (LWP 84739)]
return publisher <-------
rmw_publisher_get_actual_qos
service_name[/minimal_publisher/get_parameters], avoid_ros_namespace_conventions:0
service_name[/minimal_publisher/get_parameter_types], avoid_ros_namespace_conventions:0
service_name[/minimal_publisher/set_parameters], avoid_ros_namespace_conventions:0
service_name[/minimal_publisher/set_parameters_atomically], avoid_ros_namespace_conventions:0
service_name[/minimal_publisher/describe_parameters], avoid_ros_namespace_conventions:0
service_name[/minimal_publisher/list_parameters], avoid_ros_namespace_conventions:0
type_support(rosidl_typesupport_cpp, 0x7fdf461cf2a0, 1)
topic_name[/parameter_events], avoid_ros_namespace_conventions:0
topic_prefix[rt], topic_str[rt/parameter_events]
type_name[callbacks->message_namespace::msg::dds_::callbacks->message_name_], topic_name[/parameter_events], topic_str[rt/parameter_events]
return publisher <-------
rmw_publisher_get_actual_qos
rmw_publish----->
publish writer->write
rmw_publish return 0
topic_name[/parameter_events], avoid_ros_namespace_conventions:0
topic_prefix[rt], topic_str[rt/parameter_events]
create_publisher() ------->
type_support(rosidl_typesupport_cpp, 0x7fdf46d6d9a0, 1)
topic_name[/topic], avoid_ros_namespace_conventions:0
topic_prefix[rt], topic_str[rt/topic]
type_name[callbacks->message_namespace::msg::dds_::callbacks->message_name_], topic_name[/topic], topic_str[rt/topic]
return publisher <-------
rmw_publisher_get_actual_qos
reate_wall_timer() ------->
terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
  what():  rcl_wait() failed: read condition handle is null, at /opt/workspace/install/rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/wait.hpp:134, at /tmp/binarydeb/ros-eloquent-rcl-0.8.4/src/rcl/wait.c:633

Thread 1 "publisher_membe" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007fdf46211801 in __GI_abort () at abort.c:79
#2  0x00007fdf46866957 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007fdf4686cab6 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007fdf4686caf1 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007fdf4686caaa in std::rethrow_exception(std::__exception_ptr::exception_ptr) () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007fdf4766edd8 in rclcpp::exceptions::throw_from_rcl_error(int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, rcutils_error_state_t const*, void (*)()) () from /opt/ros/eloquent/lib/librclcpp.so
#7  0x00007fdf476746a3 in rclcpp::executor::Executor::wait_for_work(std::chrono::duration<long, std::ratio<1l, 1000000000l> >) ()
   from /opt/ros/eloquent/lib/librclcpp.so
#8  0x00007fdf476747a5 in rclcpp::executor::Executor::get_next_executable(rclcpp::executor::AnyExecutable&, std::chrono::duration<long, std::ratio<1l, 1000000000l> >) () from /opt/ros/eloquent/lib/librclcpp.so
#9  0x00007fdf47678710 in rclcpp::executors::SingleThreadedExecutor::spin() () from /opt/ros/eloquent/lib/librclcpp.so
#10 0x00007fdf47675412 in rclcpp::spin(std::shared_ptr<rclcpp::node_interfaces::NodeBaseInterface>) () from /opt/ros/eloquent/lib/librclcpp.so
#11 0x00007fdf476756cb in rclcpp::spin(std::shared_ptr<rclcpp::Node>) () from /opt/ros/eloquent/lib/librclcpp.so
#12 0x000055cca67d243e in main (argc=1, argv=0x7ffe11bbbce8) at /opt/workspace/src/examples/rclcpp/minimal_publisher/member_function.cpp:56
```

mitza-oci:
The code in `wait.hpp` requires a non-null `subscriber_info->read_condition_`.  Where should that be populated?

lij-oci:
Two places a read_condition is created:
1) in rmw_create_subscription(): read_condition = topic_reader->create_readcondition(...
2) in rmw_create_service(): read_condition = request_datareader->create_readcondition(...

rmw_opendds_cpp/src/type_support_common.hpp
adamsj-ros: @lij-oci I think we covered this. I think we can resolve this conversation since you converted to the inline function or am I mistaken?

rmw_opendds_cpp/src/rmw_publisher.cpp
adamsj-ros: @lij-oci : what's the status?

rmw_opendds_cpp/src/rmw_publisher.cpp
adamsj-ros: Based on the conversation this morning, I'm resolving this conversation. We'll move forward and ignore ROSIDL typesupport as it used in this PR.

rmw_opendds_cpp/src/rmw_publisher.cpp
adamsj-ros: I don't see a clear action here. Since this PR is about getting the manual test to pass, I'll resolve this. We can revisit form (rather than function) at a later date.

rmw_opendds_cpp/src/rmw_publisher.cpp
adamsj-ros: Yes. Are we talking about line 285 in the above?

rmw_opendds_cpp/src/type_support_common.hpp
adamsj-ros: @mitza-oci Does the most recent change work for you?

mitza-oci: I guess so?  Not really in scope here, so I'll mark it resolved.

rmw_opendds_cpp/src/process_topic_and_service_names.cpp
```diff
@@ -22,35 +22,27 @@
 
 #include "rmw_opendds_shared_cpp/namespace_prefix.hpp"
 #include "rmw_opendds_shared_cpp/opendds_include.hpp"
+#include <iostream> //?? temp
```
adamsj-ros: When committed, these kind of lines have a way of making themselves a permanent fixture or a permanent crutch to understand running code. The `RMW_SET_ERROR_MSG` should produce output to stderr. It looks like this should be coming out all the time from what I've seen in code. I'd prefer to use it instead of `cout` and `cerr` for a more permanent solution here. It's not currently working for me so I opened [this issue](https://github.com/ros2/rcutils/issues/220) to understand it better. I also recommend to learn to use gdb in our run scripts to print variables as needed.

rmw_opendds_cpp/src/process_topic_and_service_names.cpp
```diff
@@ -22,35 +22,27 @@
...
 
 bool
 _process_topic_name(
   const char * topic_name,
   bool avoid_ros_namespace_conventions,
   char ** topic_str)
 {
+  std::cout << "topic_name[" << topic_name << "], avoid_ros_namespace_conventions:" << avoid_ros_namespace_conventions << '\n'; //??
```
adamsj-ros: Same comment as for the inclusion of iostream.

rmw_opendds_cpp/src/rmw_client.cpp
```diff
@@ -101,17 +104,18 @@ rmw_create_client(
...
     goto fail;
   }
 
+  //?? get_default_publisher_qos or pass DDS::Publisher* to get_datawriter_qos
   if (!get_datawriter_qos(participant, *qos_profile, datawriter_qos)) {
     // error string was set within the function
     goto fail;
   }
-
+*/
```
adamsj-ros, line 118: What's the plan here?

rmw_opendds_cpp/src/rmw_serialize.cpp
```diff
@@ -28,7 +28,10 @@ rmw_serialize(
   const rosidl_message_type_support_t * type_support,
   rmw_serialized_message_t * serialized_message)
 {
-  RMW_OPENDDS_EXTRACT_MESSAGE_TYPESUPPORT(type_support, ts, RMW_RET_ERROR)
+  const rosidl_message_type_support_t * ts = rmw_get_message_type_support(type_support);
+  if (!ts) {
+    return NULL;
+  }
```
adamsj-ros, line 34: I thought I saw a macro to check null and print an error message. If not we should at least add an `RMW_SET_ERROR_MSG` to these type of if blocks so we know where things went bad in runtime.

rmw_opendds_cpp/src/rmw_service.cpp
```diff
@@ -101,7 +105,7 @@ rmw_create_service(
     RMW_SET_ERROR_MSG("service handle is null");
     goto fail;
   }
-
+/*
   if (!get_datareader_qos(participant, *qos_profile, datareader_qos)) {
     // error string was set within the function
     goto fail;
```
adamsj-ros, line 111: What's the plan with this commented block?

rmw_opendds_cpp/src/rmw_service.cpp
```diff
@@ -121,7 +126,7 @@ rmw_create_service(
   {
     goto fail;
   }
-
+/*
   replier = callbacks->create_replier(
     participant, request_topic_str, response_topic_str,
     &datareader_qos, &datawriter_qos,
```
adamsj-ros, line 132: What's the plan here and for the remaining commented QoS blocks?

rmw_opendds_cpp/src/rmw_subscription.cpp
```diff
@@ -205,27 +208,27 @@ rmw_create_subscription(
...
   CORBA::string_free(topic_str);
   topic_str = nullptr;
-
+/*
   if (!get_datareader_qos(participant, *qos_profile, datareader_qos)) {
     // error string was set within the function
     goto fail;
   }
-
+*/
```
adamsj-ros: What's the plan with the commented blocks in this file?

rmw_opendds_cpp/src/type_support_common.hpp
```diff
@@ -27,71 +27,56 @@
...
   const message_type_support_callbacks_t * callbacks,
   const std::string & sep)
 {
-  return
-    std::string(callbacks->message_namespace) +
-    "::" + sep + "::dds_::" + callbacks->message_name + "_";
+/*
+  return std::string(callbacks->message_namespace) + "::" + sep + "::dds_::" + callbacks->message_name + "_";
+*/
+  return std::string("callbacks->message_namespace::" + sep + "::dds_::callbacks->message_name_"); //?? temp
```
adamsj-ros, line 79: This change may get us to a running state sooner which is the goal of this PR. Are `message_namespace` and `message_name` null at this point?

PR #24, opened 2020-04-15
fixed memory leak; simplified take; removed temporary code

lij-oci:
fixed memory leak; simplified take; removed temporary code

### DIFF
--- a/rmw_opendds_cpp/src/rmw_take.cpp
+++ b/rmw_opendds_cpp/src/rmw_take.cpp
@@ -113,17 +113,6 @@ take(
   return ret;
 }
 
-rmw_ret_t
-take_serialized_message(
-  rmw_serialized_message_t * serialized_msg,
-  const rmw_subscription_t * subscription,
-  bool * taken,
-  rmw_message_info_t * message_info)
-{
-  RMW_CHECK_FOR_NULL_WITH_MSG(serialized_msg, "serialized_msg is null", return RMW_RET_ERROR);
-  return take(subscription, serialized_msg, taken, message_info);
-}
-
 extern "C"
 {
 rmw_ret_t
@@ -155,7 +144,7 @@ rmw_take_serialized_message(
   bool * taken,
   rmw_subscription_allocation_t * allocation)
 {
-  return take_serialized_message(serialized_msg, subscription, taken, nullptr);
+  return take(subscription, serialized_msg, taken, nullptr);
 }
 
 rmw_ret_t
@@ -167,7 +156,7 @@ rmw_take_serialized_message_with_info(
   rmw_subscription_allocation_t * allocation)
 {
   RMW_CHECK_FOR_NULL_WITH_MSG(message_info, "message info is null", return RMW_RET_ERROR);
-  return take_serialized_message(serialized_msg, subscription, taken, message_info);
+  return take(subscription, serialized_msg, taken, message_info);
 }
 
 rmw_ret_t
--- a/rmw_opendds_cpp/src/rmw_take.cpp
+++ b/rmw_opendds_cpp/src/rmw_take.cpp
@@ -70,9 +70,9 @@ take(
         *taken = true;
 
         if (message_info) {
-          message_info->publisher_gid->implementation_identifier = opendds_identifier;
-          memset(message_info->publisher_gid->data, 0, RMW_GID_STORAGE_SIZE);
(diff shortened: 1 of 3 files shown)


adamsj-ros (approved):

rmw_opendds_cpp/src/rmw_take.cpp
```diff
@@ -26,179 +26,114 @@
...
-    *taken = true;
-  } else {
-    *taken = false;
-  }
-
-  data_reader->return_loan(dds_messages, sample_infos);
 
-  return status == DDS::RETCODE_OK;
+  reader->return_loan(dds_messages, sample_infos);
+  return (*taken || DDS::RETCODE_NO_DATA == status) ? RMW_RET_OK : RMW_RET_ERROR;
```
mitza-oci, line 87: Before this change, wasn't NO_DATA an error?  Is this something we have to come back to later, or a deliberate change?

lij-oci: This is not a change of logic, nor a temporary code. Whether I implemented it correctly needs your verification. This c-style function has more than one values for the caller. The original code returns a bool:

  if (status == DDS::RETCODE_NO_DATA) {
    data_reader->return_loan(dds_messages, sample_infos);
    *taken = false;
    return true;
  }
  if (status != DDS::RETCODE_OK) {
    RMW_SET_ERROR_MSG("take failed");
    data_reader->return_loan(dds_messages, sample_infos);
    return false;
  }

I think *taken tells the caller whether to deserialize and then free the stream buffer. So NO_DATA is not an error, but the caller cannot process the stream. I factored out this logic from all the callers for simplicity and consistency. Its correctness needs your verification.

mitza-oci: OK, it must have just been hard to see from the diff comparison.

mitza-oci (approved):

rmw_opendds_cpp/src/rmw_take.cpp
```diff
@@ -26,179 +26,114 @@
...
-      //  break;
-      //}
+  DDS::ReturnCode_t status = reader->take(dds_messages, sample_infos, 1,
+    DDS::ANY_SAMPLE_STATE, DDS::ANY_VIEW_STATE, DDS::ANY_INSTANCE_STATE);
+
+  if (DDS::RETCODE_OK == status) {
+    DDS::SampleInfo & info = sample_infos[0];
+    if (info.valid_data) {
+      cdr_stream->buffer_length = dds_messages[0].serialized_data.length();
+      if (cdr_stream->buffer_length <= (std::numeric_limits<unsigned int>::max)()) {
```
sonndinh, line 65: The parentheses around `std::numeric_limits<unsigned int>::max` look look strange. May just remove them.

rmw_opendds_cpp/src/rmw_take.cpp
```diff
@@ -26,179 +26,114 @@
...
+        cdr_stream->buffer = reinterpret_cast<uint8_t *>(malloc(cdr_stream->buffer_length * sizeof(uint8_t)));
+        for (unsigned int i = 0; i < static_cast<unsigned int>(cdr_stream->buffer_length); ++i) {
+          cdr_stream->buffer[i] = dds_messages[0].serialized_data[i];
+        }
```
sonndinh, line 69: Can we use some utility function to copy instead of copying 1 byte at a time?

rmw_opendds_cpp/src/rmw_take.cpp
mitza-oci: It is a pattern used in some projects for compatibility with systems that have max defined as a function-like macro

rmw_opendds_cpp/src/rmw_take.cpp
```diff
@@ -26,179 +26,114 @@
...
-  return RMW_RET_OK;
+rmw_ret_t
+take_serialized_message(
+  rmw_serialized_message_t * serialized_msg,
+  const rmw_subscription_t * subscription,
+  bool * taken,
+  rmw_message_info_t * message_info)
+{
+  RMW_CHECK_FOR_NULL_WITH_MSG(serialized_msg, "serialized_msg is null", return RMW_RET_ERROR);
+  return take(subscription, serialized_msg, taken, message_info);
```
sonndinh: `take()` function already checks if `serialized_msg` is null.

sonndinh (approved):

PR #25, opened 2020-04-17
uncommented type support callback

kuznetsovmoci:


adamsj-ros:
Looks good and it works!

PR #27, opened 2020-04-28
details from test build

adamsj-ros:
Update CI to produce more interesting tests. It's still a manual process but reviewers have the ability to see running pub/sub. There's no attempt to make this to work for client/server in this PR.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -70,3 +70,4 @@ jobs:
         ros2 run examples_rclcpp_minimal_subscriber subscriber_member_function_twenty& 
         ros2 run examples_rclcpp_minimal_publisher publisher_member_function_twenty
     displayName:  run simple tests
+    
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -67,7 +67,7 @@ jobs:
         printenv|grep ROS
         export RMW_IMPLEMENTATION=rmw_opendds_cpp
         env|grep RMW
-        ros2 run examples_rclcpp_minimal_subscriber subscriber_member_function_twenty& 
+        ros2 run examples_rclcpp_minimal_subscriber subscriber_member_function_twenty 1> sub.out& 
         ros2 run examples_rclcpp_minimal_publisher publisher_member_function_twenty
+        grep "I heard" sub.out
     displayName:  run simple tests
-    
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -69,5 +69,7 @@ jobs:
         env|grep RMW
         ros2 run examples_rclcpp_minimal_subscriber subscriber_member_function_twenty 1> sub.out& 
         ros2 run examples_rclcpp_minimal_publisher publisher_member_function_twenty
+        sleep 20
+        cat sub.out
         grep "I heard" sub.out
     displayName:  run simple tests
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -53,7 +53,7 @@ jobs:
         cd $(Agent.BuildDirectory)
         source /opt/OpenDDS/setenv.sh
         source /opt/ros/eloquent/setup.bash
-        source $(Agent.BuildDirectory)/install/local_setup.bash
+        source install/local_setup.bash
         printenv|grep DDS
         printenv|grep ROS
         colcon build --symlink-install --packages-up-to rcl_interfaces examples_rclcpp_minimal_publisher examples_rclcpp_minimal_subscriber
@@ -62,7 +62,7 @@ jobs:
         cd $(Agent.BuildDirectory)
         source /opt/OpenDDS/setenv.sh
         source /opt/ros/eloquent/setup.bash
-        source $(Agent.BuildDirectory)/install/local_setup.bash
+        source install/local_setup.bash
         printenv|grep DDS
         printenv|grep ROS
         export RMW_IMPLEMENTATION=rmw_opendds_cpp
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -42,6 +42,7 @@ jobs:
         ls s/
     displayName:  rmw build repo and dependencies
   - script: |
+        cd $(Agent.BuildDirectory)
         source /opt/OpenDDS/setenv.sh
         source /opt/ros/eloquent/setup.bash
         printenv|grep DDS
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -23,17 +23,50 @@ jobs:
...
+        source $(Agent.BuildDirectory)/install/local_setup.bash
         printenv|grep DDS
         printenv|grep ROS
-        colcon build --symlink-install
-    displayName:  environment setup and build
+        export RMW_IMPLEMENTATION=rmw_opendds_cpp
+        env|grep RMW
+        ros2 run examples_rclcpp_minimal_subscriber subscriber_member_function_twenty& 
+        ros2 run examples_rclcpp_minimal_publisher publisher_member_function_twenty
+    displayName:  run simple tests
```
mitza-oci: Newline is missing

azure-pipelines.yml
```diff
@@ -23,17 +23,50 @@ jobs:
...
+        source $(Agent.BuildDirectory)/install/local_setup.bash
+        printenv|grep DDS
+        printenv|grep ROS
+        colcon build --symlink-install --packages-up-to rcl_interfaces examples_rclcpp_minimal_publisher examples_rclcpp_minimal_subscriber
+    displayName:  build simple tests
+  - script: |
+        cd $(Agent.BuildDirectory)
         source /opt/OpenDDS/setenv.sh
         source /opt/ros/eloquent/setup.bash
+        source $(Agent.BuildDirectory)/install/local_setup.bash
```
mitza-oci:
```suggestion
        source install/local_setup.bash
```
Not necessarily wrong... (also on 56)

mitza-oci (changes requested):

azure-pipelines.yml
adamsj-ros: The $(Agent.BuildDirectory) is required because of the layout of the workspace. Repos are cloned into $(Agent.BuildDirectory)/s and the build scripts `rmw_build` are cloned to $(Agent.BuildDirectory). The build scripts expect these relative paths.

kuznetsovmoci (commented):
New line at the end of the file is missing.
Also (non critical), is there an easy way to avoid changing directories (cd $(Agent.BuildDirectory)) and specify $(Agent.BuildDirectory) in the command paths?

adamsj-ros:
@mitza-oci and @lij-oci: Added a redirect subscriber output to a file and then grep for "I heard" as you suggested which failed the first time I ran it because the file was empty. I had to add a sleep to make sure the file was there because it's taking awhile for the subscriber to exit. The publisher seems like it exits quickly. I suspect [this line](https://dev.azure.com/rmw-opendds/rmw_build/_build/results?buildId=402&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=b525c197-f769-5e52-d38a-e6301f5260f2&l=46) as being a clue that the destruction could be cleaner.

azure-pipelines.yml
mitza-oci: My point is that it's already your working directory

azure-pipelines.yml
```diff
@@ -23,17 +23,53 @@ jobs:
...
         printenv|grep DDS
         printenv|grep ROS
-        colcon build --symlink-install
-    displayName:  environment setup and build
+        export RMW_IMPLEMENTATION=rmw_opendds_cpp
+        env|grep RMW
+        ros2 run examples_rclcpp_minimal_subscriber subscriber_member_function_twenty 1> sub.out&
```
mitza-oci, line 70: This script puts a process in the background without waiting for it.  Maybe that would be a better approach than the `sleep` that's 2 lines down.

mitza-oci (changes requested):

azure-pipelines.yml
lij-oci: Would it be cleaner if pub/sub exit by themselves? For example, pub sends n messages and quits. Sub receives <=n messages and quits.

mitza-oci: That would be good eventually, but we may not be ready for it yet.
Until then we can use external process control like wait and kill

adamsj-ros: To be clear, these particular pub and sub applications exit automatically after 20 seconds. I have to background the subscriber so it can run concurrently with the publisher otherwise we won't see any messages received in the subscriber. The sleep 20 is required because for some reason the subscriber is taking many seconds to shutdown and make the prints available in sub.out.

mitza-oci: Right, so if you'd use a "wait" in the shell then we'd know that the background process is done and doesn't still have an open file handle (and data possibly buffered) for the output file.

adamsj-ros: Much better. Thanks. I wasn't aware of the `wait` command.

mitza-oci (approved):
I still like relative paths for "source" but I guess that's down to personal preference

adamsj-ros:
> is there an easy way to avoid changing directories (cd $(Agent.BuildDirectory))

The nature of this CI is that it automatically clones this repo into a particular path which happens to be `$(Agent.BuildDirectory)/s`. This became the template for my workspace in the CI meaning that the source repos are cloned into `$(Agent.BuildDirectory)/s` and `rmw_build` is cloned to a level higher in `$(Agent.BuildDirectory)` much like the layout of our dev workspaces. This allows all the scripts to run correctly in maintaining the correct relative path structure. The side effect is that I need to treat `$(Agent.BuildDirectory)` as the root of the workspace so some operations need to be run from there. That's why you see all the references to Agent.BuildDirectory.

mitza-oci:
> > is there an easy way to avoid changing directories (cd $(Agent.BuildDirectory))
> 
> The nature of this CI is that it automatically clones this repo into a particular path which happens to be `$(Agent.BuildDirectory)/s`. This became the template for my workspace in the CI meaning that the source repos are cloned into `$(Agent.BuildDirectory)/s` and `rmw_build` is cloned to a level higher in `$(Agent.BuildDirectory)` much like the layout of our dev workspaces. This allows all the scripts to run correctly in maintaining the correct relative path structure. The side effect is that I need to treat `$(Agent.BuildDirectory)` as the root of the workspace so some operations need to be run from there. That's why you see all the references to Agent.BuildDirectory.

If you'd like to remove some references to Agent.BuildDirectory, see my note about using relative paths.

adamsj-ros:
> I still like relative paths for "source" but I guess that's down to personal preference

I see what you're saying now. I made the change to cd into that directory and source the files in different parts of the development so I forgot that the absolute paths were not longer required. I agree with the reduced complexity of using relative paths.

PR #28, opened 2020-05-08
rmw client and service

lij-oci:


rmw_opendds_cpp/include/rmw_opendds_cpp/opendds_static_service_info.hpp
```diff
@@ -23,10 +23,12 @@ extern "C"
 {
 struct OpenDDSStaticServiceInfo
 {
+  const service_type_support_callbacks_t * callbacks_;
   void * replier_;
-  DDS::DataReader_var request_datareader_;
+  DDS::DataReader_var request_reader_;
   DDS::ReadCondition_var read_condition_;
-  const service_type_support_callbacks_t * callbacks_;
+  OpenDDSStaticServiceInfo(const service_type_support_callbacks_t * callbacks)
```
mitza-oci:
```suggestion
  explicit OpenDDSStaticServiceInfo(const service_type_support_callbacks_t * callbacks)
```

rmw_opendds_cpp/src/rmw_service.cpp
```diff
@@ -26,319 +26,241 @@
...
-  const service_type_support_callbacks_t * callbacks =
-    static_cast<const service_type_support_callbacks_t *>(type_support->data);
-  if (!callbacks) {
-    RMW_SET_ERROR_MSG("callbacks handle is null");
-    return NULL;
+  RMW_CHECK_FOR_NULL_WITH_MSG(service_name, "service_name is null", return nullptr);
+  CORBA::String_var request_topic;
+  CORBA::String_var response_topic;
+  if (!_process_service_name(service_name, qos_profile->avoid_ros_namespace_conventions,
+      &request_topic.out(), &response_topic.out())) {
```
mitza-oci: I'm not 100% sure this use of `.out` is going to work.  Since `_process_service_name` is internal we could change it's argument types to be friendlier (for example, use std::string).

rmw_opendds_cpp/src/rmw_service.cpp
```diff
@@ -26,319 +26,241 @@
...
+    RMW_SET_ERROR_MSG("failed to set_default_datareader_qos");
+    return nullptr;
   }
 
-  dds_subscriber = request_datareader->get_subscriber();
-  status = participant->get_default_subscriber_qos(subscriber_qos);
-  if (status != DDS::RETCODE_OK) {
-    RMW_SET_ERROR_MSG("failed to get default subscriber qos");
-    goto fail;
+  // create service
```
mitza-oci: It is pretty clear what is happening based on the next line, so I'd remove this comment.

rmw_opendds_cpp/src/rmw_service.cpp
```diff
@@ -26,319 +26,241 @@
...
   }
+  void * buf = nullptr;
+  try {
+    service->implementation_identifier = opendds_identifier;
+    service->data = nullptr;
+    service->service_name = reinterpret_cast<const char *>(rmw_allocate(strlen(service_name) + 1));
+    if (!service->service_name) {
+      throw std::string("failed to allocate memory for service name");
+    }
+    memcpy(const_cast<char*>(service->service_name), service_name, strlen(service_name) + 1);
```
mitza-oci: This is strcpy

rmw_opendds_cpp/src/rmw_service.cpp
```diff
@@ -26,319 +26,241 @@
...
 
-  dds_publisher = response_datawriter->get_publisher();
-  status = participant->get_default_publisher_qos(publisher_qos);
-  if (status != DDS::RETCODE_OK) {
-    RMW_SET_ERROR_MSG("failed to get default subscriber qos");
-    goto fail;
-  }
+    // create OpenDDSStaticServiceInfo
```
mitza-oci: It is pretty clear what is happening based on the next line, so I'd remove this comment.

mitza-oci (changes requested):

rmw_opendds_cpp/src/rmw_service.cpp
lij-oci: done

rmw_opendds_cpp/src/rmw_service.cpp
lij-oci: done

rmw_opendds_cpp/include/rmw_opendds_cpp/opendds_static_service_info.hpp
lij-oci: done

rmw_opendds_cpp/src/rmw_service.cpp
lij-oci: done

mitza-oci:
@lij-oci this PR's title is "rmw service".  Do you want the client changes here (if so, change the title), or do you want to keep them separate?

@mitza-oci Yeah, you are right. Since service and client share the same rosidl_service_type_support_t and some helper functions, changing one often needs to change the other. So I will change the title to "rmw client and service".

rmw_opendds_cpp/src/process_topic_and_service_names.cpp
```diff
@@ -23,54 +23,22 @@
...
 
-bool
-_process_topic_name(
-  const char * topic_name,
+void
+get_topic_name(
+  const std::string topic_name,
   bool avoid_ros_namespace_conventions,
-  char ** topic_str)
+  std::string& topic_str)
```
mitza-oci: this looks like the return value, I suggest making it the actual return value

rmw_opendds_cpp/src/process_topic_and_service_names.cpp
```diff
@@ -23,54 +23,22 @@
 #include "rmw_opendds_shared_cpp/namespace_prefix.hpp"
 #include "rmw_opendds_shared_cpp/opendds_include.hpp"
 
-bool
-_process_topic_name(
-  const char * topic_name,
+void
+get_topic_name(
+  const std::string topic_name,
```
mitza-oci:
```suggestion
  const std::string& topic_name,
```

mitza-oci (changes requested):

mitza-oci:
@lij-oci I posted a comment here yesterday, please reply

rmw_opendds_cpp/src/process_topic_and_service_names.cpp
```diff
@@ -23,54 +24,27 @@
...
-    *topic_str = CORBA::string_dup(concat_str);
-    allocator.deallocate(concat_str, allocator.state);
-  } else {
-    RMW_SET_ERROR_MSG("could not allocate memory for topic string");
-    success = false;
+  if (!topic_name || strlen(topic_name) == 0) {
+    return "";
+  }
+  if (avoid_ros_namespace_conventions) {
+    return topic_name;
```
adamsj-ros, line 36: Since the topic_name is passed by pointer, I'm wondering if it wouldn't be better to return an error code instead.

rmw_opendds_cpp/src/rmw_subscription.cpp
```diff
@@ -167,25 +167,21 @@ rmw_create_subscription(
   //  goto fail;
   //}
 
-  if (!topic_name || strlen(topic_name) == 0) {
+  std::string topic_str = get_topic_str(topic_name, qos_profile->avoid_ros_namespace_conventions);
+  if (topic_str.empty()) {
```
adamsj-ros, line 171: back to the error code from the called function. A string empty here doesn't really convey a lot of meaning.

rmw_opendds_cpp/src/rmw_service.cpp
```diff
@@ -26,319 +26,236 @@
...
 // This affects code in this file, but there is a similar variable in:
 //   rmw_opendds_shared_cpp/shared_functions.cpp
 // #define DISCOVERY_DEBUG_LOGGING 1
 
 extern "C"
 {
+rmw_ret_t clean_service(OpenDDSNodeInfo & node_info, rmw_service_t * service)
+{
+  auto ret = RMW_RET_OK;
+  if (!service) {
```
adamsj-ros, line 41: Have we adopted a convention for the standard nullptr check? Would it be better to check if it's equal to `nullptr` instead of this method?

rmw_opendds_cpp/src/rmw_client.cpp
```diff
@@ -27,309 +27,233 @@
...
+#include <dds/DCPS/Marked_Default_Qos.h>
+
 // Uncomment this to get extra console output about discovery.
 // This affects code in this file, but there is a similar variable in:
 //   rmw_opendds_shared_cpp/shared_functions.cpp
 // #define DISCOVERY_DEBUG_LOGGING 1
 
 extern "C"
 {
+rmw_ret_t clean_client(OpenDDSNodeInfo & node_info, rmw_client_t * client)
```
adamsj-ros, line 39: Not that it matters a whole lot, but I'm wondering why you pass by reference and by pointer? I think it would be good to remain consistent. That way it's easier to know if you need to dereference or not the function parameters.

rmw_opendds_cpp/src/rmw_client.cpp
```diff
@@ -27,309 +27,233 @@
...
+{
+  auto ret = RMW_RET_OK;
+  if (!client) {
```
adamsj-ros, line 42: I suggest to check against `nullptr` instead of not true. I'm not sure that it can be guaranteed by all compilers that this check will work if the pointer is set to `nullptr` as it should be set. If agreed, need to make changes to the the reset of this review.

rmw_opendds_cpp/src/rmw_client.cpp
```diff
@@ -27,309 +27,233 @@
...
+    return ret;
+  }
+  auto info = static_cast<OpenDDSStaticClientInfo*>(client->data);
+  if (info) {
```
adamsj-ros, line 46: check `nullptr`

rmw_opendds_cpp/src/rmw_client.cpp
```diff
@@ -27,309 +27,233 @@
...
   const rmw_qos_profile_t * qos_profile)
 {
-  if (!node) {
-    RMW_SET_ERROR_MSG("node is null");
-    return NULL;
-  }
-  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
-    node handle, node->implementation_identifier, opendds_identifier,
-    return NULL)
+  RMW_CHECK_FOR_NULL_WITH_MSG(node, "node is null", return nullptr);
```
adamsj-ros, line 96: This check is explicitly checking against `NULL`. I'm wondering if it's correct if using `nullptr`.

rmw_opendds_cpp/src/rmw_client.cpp
```diff
@@ -27,309 +27,233 @@
...
-    return NULL;
-  }
-  auto participant = static_cast<DDS::DomainParticipant *>(node_info->participant);
-  if (!participant) {
-    RMW_SET_ERROR_MSG("participant is null");
-    return NULL;
-  }
+  RMW_CHECK_FOR_NULL_WITH_MSG(service_name, "service_name is null", return nullptr);
+  std::string request_topic;
+  std::string response_topic;
```
adamsj-ros, line 112: init strings?

rmw_opendds_cpp/src/rmw_client.cpp
```diff
@@ -27,309 +27,233 @@
...
+    std::strcpy(const_cast<char*>(client->service_name), service_name);
 
-  // update attached subscriber
-  dds_subscriber->set_qos(subscriber_qos);
+    buf = rmw_allocate(sizeof(OpenDDSStaticClientInfo));
+    if (!buf) {
+      throw std::string("failed to allocate memory for client info");
+    }
+    OpenDDSStaticClientInfo * info = nullptr;
+    RMW_TRY_PLACEMENT_NEW(info, buf, throw 1, OpenDDSStaticClientInfo,
```
adamsj-ros, line 171: Is `throw 1` meaningful enough?

rmw_opendds_cpp/src/rmw_client.cpp
iguessthislldo: [I think C++03 didn't have this guarantee but C++11 does.](https://en.cppreference.com/w/cpp/language/implicit_conversion#Boolean_conversions)

mitza-oci: C++03 does have the guarantee, as does C.  Unless I'm misunderstanding you.

mitza-oci:
> @lij-oci this PR's title is "rmw service". Do you want the client changes here (if so, change the title), or do you want to keep them separate?
> 
> @mitza-oci Yeah, you are right. Since service and client share the same rosidl_service_type_support_t and some helper functions, changing one often needs to change the other. So I will change the title to "rmw client and service".

Please don't edit github comments to reply.  It doesn't send email notifications for those.

rmw_opendds_cpp/src/rmw_client.cpp
lij-oci: std::string::string() will create an empty string.

rmw_opendds_cpp/src/rmw_client.cpp
lij-oci: The purpose of clean_client() is to clear (if client is unnull) client's members and then free the client, so the client is passed as a point. In order to do so, node needs to do some housekeeping: remove_information and trigger_graph_guard_condition. That's why node is passed as a mutable reference, not a const reference, nor a point (since node is not freed here).

rmw_opendds_cpp/src/process_topic_and_service_names.cpp
lij-oci: This helper function is to decorate the topic_name. If it's null or empty, an empty string is returned so the caller can take an appropriate action (return RMW_RET_ERROR or nullptr).

rmw_opendds_cpp/src/rmw_subscription.cpp
lij-oci: That's true. It only tells the caller that there is no valid topic_name, and that should be sufficient. Otherwise, we have to pass string& to get the decorated topic_name and return an error code or bool, and the caller has to check the return code before using the value in the string&.

rmw_opendds_cpp/src/rmw_client.cpp
adamsj-ros: I thought I had read that `nullptr` value could be different for different compilers. I guess that doesn't mean that a boolean check like this wouldn't work.

rmw_opendds_cpp/src/rmw_client.cpp
adamsj-ros: Talked during standup and confirmed throw 1 is appropriate.

rmw_opendds_cpp/src/rmw_client.cpp
iguessthislldo: > I thought I had read that nullptr value could be different for different compilers. I guess that doesn't mean that a boolean check like this wouldn't work.

I remember reading at some point that `NULL` could be something else other than `0` on some (probably historical) platforms, so maybe that's what you're thinking of?

adamsj-ros:
@mitza-oci Do you want me to follow up with your requested changes or are you going to have an opportunity to check them out?

mitza-oci:
> @mitza-oci Do you want me to follow up with your requested changes or are you going to have an opportunity to check them out?

I think a lot has changed on this branch since I reviewed (not just responses to reviews but increased scope).  So I'll leave it up to you to determine if you want to go ahead and merge it.

adamsj-ros:
> So I'll leave it up to you to determine if you want to go ahead and merge it

@lij-oci Since you're intimately familiar with this code, could you go back through the items marked resolved above and make sure you have a meaningful comment on how you resolved it. I'm looking for a little more than "done" if possible. Especially important and helpful for me would be to add responses to _all_ Adam's comments for these items now marked resolved. Thanks.

lij-oci:
@adamsj-oci, @mitza-oci Those earlier few changes I wrote "done" are simple changes and Adam advised me to use the resolved button instead of writing "done". Now the important thing here is that I would like Adam to review rmw_client.cpp. It's similar to rmw_service, and shares the same type support interface. They should be merged together to make this work.

rmw_opendds_cpp/src/rmw_client.cpp
```diff
@@ -27,309 +27,233 @@
...
+        info->read_condition_ = nullptr;
+      }
+      info->response_reader_ = nullptr;
+    } else if (info->read_condition_) {
+      RMW_SET_ERROR_MSG("cannot delete readcondition because the datareader is null");
+      ret = RMW_RET_ERROR;
+    }
+
+    if (info->callbacks_ && info->requester_) {
+      DDS::DataWriter_var writer = static_cast<DDS::DataWriter *>(
```
mitza-oci, line 65: I don't think there's anything technically wrong here, but I think this is harder to read/maintain/verify than it should be.  In https://github.com/oci-labs/rosidl_typesupport_opendds/pull/18 some types were changed from `void*` to their actual DDS API types, but these were not.  I think we should do a follow-on PR in that repo to make this more consistent.

rmw_opendds_cpp/src/rmw_client.cpp
```diff
@@ -27,309 +27,233 @@
...
   }
+  void * buf = nullptr;
+  try {
+    client->implementation_identifier = opendds_identifier;
+    client->data = nullptr;
+    client->service_name = reinterpret_cast<const char *>(rmw_allocate(strlen(service_name) + 1));
+    if (!client->service_name) {
+      throw std::string("failed to allocate memory for service name");
+    }
+    std::strcpy(const_cast<char*>(client->service_name), service_name);
```
mitza-oci:
```suggestion
    std::strcpy(const_cast<char *>(client->service_name), service_name);
```

rmw_opendds_cpp/src/rmw_client.cpp
mitza-oci: I think we should revisit these later.  There are other instances of it from the pub/sub functions.

rmw_opendds_cpp/src/rmw_client.cpp
```diff
@@ -27,309 +27,233 @@
...
 
-  // update attached subscriber
-  dds_subscriber->set_qos(subscriber_qos);
+    buf = rmw_allocate(sizeof(OpenDDSStaticClientInfo));
+    if (!buf) {
+      throw std::string("failed to allocate memory for client info");
+    }
+    OpenDDSStaticClientInfo * info = nullptr;
+    RMW_TRY_PLACEMENT_NEW(info, buf, throw 1, OpenDDSStaticClientInfo,
+      static_cast<const service_type_support_callbacks_t*>(type_support->data))
```
mitza-oci:
```suggestion
      static_cast<const service_type_support_callbacks_t *>(type_support->data))
```

rmw_opendds_cpp/src/rmw_client.cpp
```diff
@@ -27,309 +27,233 @@
...
-    goto fail;
-  }
-  // Use a placement new to construct the OpenDDSStaticClientInfo in the preallocated buffer.
-  RMW_TRY_PLACEMENT_NEW(client_info, buf, goto fail, OpenDDSStaticClientInfo, )
-  buf = nullptr;  // Only free the client_info pointer; don't need the buf pointer anymore.
-  client_info->requester_ = nullptr; //requester;
-  client_info->callbacks_ = nullptr; //callbacks;
-  client_info->response_datareader_ = nullptr; //response_datareader;
-  client_info->read_condition_ = nullptr; //read_condition;
+    info->response_reader_ = static_cast<DDS::DataReader*>(
```
mitza-oci:
```suggestion
    info->response_reader_ = static_cast<DDS::DataReader *>(
```

rmw_opendds_cpp/src/rmw_client.cpp
```diff
@@ -27,309 +27,233 @@
...
-    rmw_client_free(client);
-  }
-  if (response_datareader) {
-    if (dds_subscriber->delete_datareader(response_datareader) != DDS::RETCODE_OK) {
-      std::stringstream ss;
-      ss << "leaking datareader while handling failure at " <<
-        __FILE__ << ":" << __LINE__ << '\n';
-      (std::cerr << ss.str()).flush();
+    // update node_info publisher_listener
+    DDS::DataWriter_var writer = static_cast<DDS::DataWriter*>(
```
mitza-oci:
```suggestion
    DDS::DataWriter_var writer = static_cast<DDS::DataWriter *>(
```

rmw_opendds_cpp/src/rmw_service.cpp
lij-oci: used std::string

rmw_opendds_cpp/src/rmw_client.cpp
```diff
@@ -27,309 +27,233 @@
...
+      info->callbacks_->get_request_datawriter(info->requester_));
+    if (!writer) {
+      throw std::string("reply writer is null");
     }
+    DDS::Topic_ptr wtopic = writer->get_topic();
```
mitza-oci:
```suggestion
    DDS::Topic_var wtopic = writer->get_topic();
```

rmw_opendds_cpp/src/rmw_service.cpp
```diff
@@ -26,319 +26,236 @@
...
   }
+  void * buf = nullptr;
+  try {
+    service->implementation_identifier = opendds_identifier;
+    service->data = nullptr;
+    service->service_name = reinterpret_cast<const char *>(rmw_allocate(strlen(service_name) + 1));
+    if (!service->service_name) {
+      throw std::string("failed to allocate memory for service name");
+    }
+    std::strcpy(const_cast<char*>(service->service_name), service_name);
```
mitza-oci:
```suggestion
    std::strcpy(const_cast<char *>(service->service_name), service_name);
```

rmw_opendds_cpp/src/rmw_service.cpp
```diff
@@ -26,319 +26,236 @@
...
-    RMW_SET_ERROR_MSG("failed to get default subscriber qos");
-    goto fail;
-  }
+    buf = rmw_allocate(sizeof(OpenDDSStaticServiceInfo));
+    if (!buf) {
+      throw std::string("failed to allocate memory for service info");
+    }
+    OpenDDSStaticServiceInfo * info = nullptr;
+    RMW_TRY_PLACEMENT_NEW(info, buf, throw 1, OpenDDSStaticServiceInfo,
+      static_cast<const service_type_support_callbacks_t*>(type_support->data))
```
mitza-oci:
```suggestion
      static_cast<const service_type_support_callbacks_t *>(type_support->data))
```

mitza-oci (changes requested):

rmw_opendds_cpp/src/process_topic_and_service_names.cpp
adamsj-ros: Looks like in the final version, `topic_name` became a `const char *`. Why?

rmw_opendds_cpp/src/rmw_client.cpp
adamsj-ros: I created issue #30 to capture this for now.

rmw_opendds_cpp/src/rmw_client.cpp
lij-oci: @mitza-oci Good catch. I also changed Line 198: DDS::TopicDescription_ptr to _var.

rmw_opendds_cpp/src/process_topic_and_service_names.cpp
lij-oci: Because we eliminated the external checks, and saved unnecessary temporary std::string creation on the call stack and in concatenations. Here, we concatenate only when necessary. We still have to create the return std::string. Depending on the compiler setting, we may not have a copy overhead if c++17 (copy elision) is used.

PR #35, opened 2020-06-01
rmw_wait

lij-oci:
# Description
Implemented rmw_wait().

Fixes (Trello link here OR reason to fix if not on Trello)
https://trello.com/c/pzVotWi6/50-2-implement-rmwwait-and-wait

## Type of change
- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?
Code review. Pub/sub minimal examples should run smoothly.

# Checklist:
- New and existing system tests pass locally with my changes (when applicable)

### DIFF
--- a/rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/wait.hpp
+++ b/rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/wait.hpp
@@ -104,7 +104,6 @@ wait(
   }
 
   // add a condition for each service
-/* TODO: uncommnet this block when type support is ready.
   if (services) {
     for (size_t i = 0; i < services->service_count; ++i) {
       auto info = static_cast<ServiceInfo *>(services->services[i]);
@@ -114,7 +113,10 @@ wait(
       }
       if (!info->read_condition_) {
         RMW_SET_ERROR_MSG("read condition is null");
+        break;
+/* TODO: delete the line above and uncomment the line below when service typesupport is ready.
         return RMW_RET_ERROR;
+*/
       }
       rmw_ret_t ret = check_attach_condition_error(dds_wait_set.attach_condition(info->read_condition_));
       if (ret != RMW_RET_OK) {
@@ -122,7 +124,6 @@ wait(
       }
     }
   }
-*/
 
   // add a condition for each client
   if (clients) {
@@ -173,11 +174,6 @@ wait(
       if (!(j < active_conditions.length())) {
         subscriptions->subscribers[i] = nullptr;
       }
-
-      if (dds_wait_set.detach_condition(info->read_condition_) != DDS::RETCODE_OK) {
-        RMW_SET_ERROR_MSG("failed to detach condition");
-        return RMW_RET_ERROR;
-      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/wait.hpp
```diff
@@ -36,344 +36,238 @@ wait(
...
-          RMW_SET_ERROR_MSG("Failed to get detach condition from wait set");
+  // Cleaner destructor will clean up the wait set (previously done in wait_set destructor)
+  struct Cleaner
+  {
+    OpenDDSWaitSetInfo & info;
+    DDS::WaitSet & waitset;
+    Cleaner(OpenDDSWaitSetInfo & i, DDS::WaitSet & w) : info(i), waitset(w) {}
+    ~Cleaner()
+    {
+      auto attached = static_cast<DDS::ConditionSeq *>(info.attached_conditions);
```
mitza-oci: Why is this cast necessary?  Looks like it already has the type it's being cast to.

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/wait.hpp
```diff
@@ -36,344 +36,238 @@ wait(
...
+      rmw_ret_t ret = check_attach_condition_error(dds_wait_set->attach_condition(condition));
+      if (ret != RMW_RET_OK) {
+        return ret;
       }
     }
   }
-/*  TODO: uncommnet this block when type support is ready and service is properly implemented.
+
   // add a condition for each service
+/* TODO: uncommnet this block when type support is ready.
```
mitza-oci: Can this be un-commented now?

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/wait.hpp
```diff
@@ -36,344 +36,238 @@ wait(
...
+        RMW_SET_ERROR_MSG("failed to detach condition");
         return RMW_RET_ERROR;
       }
     }
   }
-/*  TODO: uncommnet this block when type support is ready and service is properly implemented.
-  // set service handles to zero for all not triggered conditions
+
+  // reset service for all untriggered conditions
+/* TODO: uncommnet this block when type support is ready.
```
mitza-oci: un-comment?

mitza-oci (changes requested):

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/wait.hpp
lij-oci: It will crash CI tests.

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/wait.hpp
lij-oci: It will crash CI tests.

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/wait.hpp
mitza-oci: This is the same idea as the previous PR.  You can use runtime control flow to have it skip the block of code that's impacted.  Then the code we are reviewing is at least compiled, if not yet tested.

adamsj-ros (approved):

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/wait.hpp
adamsj-ros: I don't think we are opposed to merging if there is a test issue as long as it can be explained or handled in a temporary way as Adam had suggested. The problem right now is that we only really have one test and breaking that one puts us in the blind if we accept it.

mitza-oci (approved):

PR #37, opened 2020-06-12
Uncommented use of type support

kuznetsovmoci:


adamsj-ros:
Everything looks good. I will merge this and then we'll rerun the integration test in my draft PR. If we don't have any failures then we can call it good. If we find issues, then we'll have to assess those issues.

PR #51, opened 2020-07-17
Interop fixes pub/sub

adamsj-ros:
It's been observed that other DDS/RTPS RMW implementations are using 20 bytes to transmit our base test while our OpenDDS RMW is using 64 bytes. At least one part of the interop issues are to match the serialization for user payload between OpenDDS RMW and FastRTPS/Cyclone.

### DIFF
--- a/rmw_opendds_cpp/src/rmw_publisher.cpp
+++ b/rmw_opendds_cpp/src/rmw_publisher.cpp
@@ -32,6 +32,7 @@
 // include patched generated code from the build folder
 //#include "opendds_static_serialized_dataTypeSupportC.h"
 #include "opendds_static_serialized_dataTypeSupportImpl.h"
+#include "dds/DCPS/DataWriterImpl_T.h"
 
 // Uncomment this to get extra console output about discovery.
 // This affects code in this file, but there is a similar variable in:
@@ -211,6 +212,10 @@ rmw_create_publisher(
     if (!publisher_info->topic_writer_) {
       throw std::string("failed to create datawriter");
     }
+
+    OpenDDS::DCPS::DataWriterImpl_T<OpenDDSStaticSerializedData> * trt = dynamic_cast<OpenDDS::DCPS::DataWriterImpl_T<OpenDDSStaticSerializedData>*>(publisher_info->topic_writer_.in());
+
+    trt->set_marshal_skip_serialize(true);
 
     // set remaining OpenDDSStaticPublisherInfo members
     publisher_info->callbacks_ = callbacks;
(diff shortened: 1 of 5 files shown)


kuznetsovmoci:
Will it make it incompatible with RTI?

kuznetsovmoci (commented):
Will it make it incompatible with RTI?

rmw_opendds_cpp/src/rmw_publisher.cpp
```diff
@@ -212,6 +213,10 @@ rmw_create_publisher(
       throw std::string("failed to create datawriter");
     }
 
+    OpenDDS::DCPS::DataWriterImpl_T<OpenDDSStaticSerializedData> * trt = dynamic_cast<OpenDDS::DCPS::DataWriterImpl_T<OpenDDSStaticSerializedData>*>(publisher_info->topic_writer_.in());
+
+    trt->set_marshal_skip_serialize(true);
```
jwillemsen, line 218: Wouldn't this be a candidate for a QoS extension in order to not use internal OpenDDS classes?

adamsj-ros:
> Will it make it incompatible with RTI?

I'm not sure. I'm more concerned that it's compatible with the default RMW -- eProsima's FastRTPS. It's compatible with FastRTPS and CycloneDDS. If RTI doesn't work with FastRTPS and Cyclone then I think there are other issues. 

Keep in mind this is only for pub/sub. I checked RPC and there are big issues with FastRTPS but FastRTPS apparently has big issues with Cyclone too for RPC. We are close to compatible with Cyclone in RPC but that's not the scope of this PR.

rmw_opendds_cpp/src/rmw_publisher.cpp
adamsj-ros: Could you explain what you mean a little more? I like the idea of not using OpenDDS namespace items in the RMW but we have about a dozen source files doing so at the moment. In other words, I'm not sure how important that is as a design goal in the RMW.

kuznetsovmoci (approved):

rmw_opendds_cpp/src/rmw_publisher.cpp
jwillemsen: Could this be done with an extension of one of the QoS structs so that you can specify a custom QoS in your code to enable this internal feature? Not sure how generic this all can be used.

mitza-oci: This should not be used anywhere.  Even in RMW, it's only there as a work around.

PR #53, opened 2021-01-18
RmwAllocateFree

lij-oci:
This template class provides consistent rmw allocate and free. It is used as a step in fixing the errors in rmw object creation and destruction.

This class is to be verified by inspection/code review, and later used in other objects including node, publisher, subscriber, client, and service.


rmw_opendds_cpp/src/rmw_init.cpp
```diff
@@ -25,14 +25,15 @@ extern "C"
...
-  RCUTILS_CHECK_ALLOCATOR(&allocator, return RMW_RET_INVALID_ARGUMENT);
-  if (NULL != init_options->implementation_identifier) {
+  if (!is_zero_initialized(init_options)) {
     RMW_SET_ERROR_MSG("expected zero-initialized init_options");
     return RMW_RET_INVALID_ARGUMENT;
   }
+  RCUTILS_CHECK_ALLOCATOR(&allocator, return RMW_RET_INVALID_ARGUMENT);
   init_options->instance_id = 0;
   init_options->implementation_identifier = opendds_identifier;
+  init_options->domain_id = 123; //?? RMW_DEFAULT_DOMAIN_ID //??added
```
adamsj-ros: Where did this magic number come from? Is it not set when `init_options` is passed in?

rmw_opendds_cpp/src/rmw_init.cpp
```diff
@@ -42,13 +43,12 @@ rmw_ret_t
 rmw_init_options_copy(const rmw_init_options_t * src, rmw_init_options_t * dst)
 {
   RMW_CHECK_ARGUMENT_FOR_NULL(src, RMW_RET_INVALID_ARGUMENT);
-  RMW_CHECK_ARGUMENT_FOR_NULL(dst, RMW_RET_INVALID_ARGUMENT);
```
adamsj-ros, line 45: why remove the dst check?

kuznetsovmoci (approved):

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/RmwAllocateFree.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+  // managed create and destroy
+  typedef std::unique_ptr<T, Deleter> Ptr;
+
+  static Ptr createPtr()
+  {
+    return Ptr(create());
+  }
+
+  template<typename ... Args>
```
adamsj-ros, line 106: I think I understand why we would use ellipses (variable number of args) but I'm concerned that this could cause problems down the road especially if functional safety is ever a concern. Is there another way to do this?

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/RmwAllocateFree.hpp
```diff
@@ -0,0 +1,140 @@
...
+};
+
+struct RmwStr
+{
+  static char * cpy(const char * str)
+  {
+    if (str) {
+      char * t = reinterpret_cast<char *>(rmw_allocate(sizeof(char) * strlen(str) + 1));
+      if (t) {
+        memcpy(t, str, strlen(str) + 1);
```
adamsj-ros, line 120: would we not want to use strcpy instead of memcpy here?

rmw_opendds_shared_cpp/src/init.cpp
```diff
@@ -13,17 +13,68 @@
...
+shutdown(rmw_context_t& context)
+{
+  if (context.impl) {
+    RmwAllocateFree<rmw_context_impl_t>::destroy(context.impl);
   }
-  TheServiceParticipant->set_default_discovery(OpenDDS::DCPS::Discovery::DEFAULT_RTPS);
   return RMW_RET_OK;
 }
+
+bool is_zero_initialized(const rmw_init_options_t * o)
```
adamsj-ros: could we use a more meaningful variable name than `o`?

rmw_opendds_shared_cpp/src/init.cpp
```diff
@@ -13,17 +13,68 @@
...
+{
+  return o && o->instance_id == 0
+    && !(o->implementation_identifier)
+    && o->domain_id == RMW_DEFAULT_DOMAIN_ID
+    //?? RMW_SECURITY_ENFORCEMENT_PERMISSIVE in rmw_get_default_security_options()
```
adamsj-ros: Remove comment?

rmw_opendds_shared_cpp/src/init.cpp
```diff
@@ -13,17 +13,68 @@
...
+    && o->security_options.enforce_security == 0 && !(o->security_options.security_root_path)
+    && o->localhost_only == RMW_LOCALHOST_ONLY_DEFAULT
+    && !(o->enclave)
+    && !(o->impl);
+}
+
+bool is_zero_initialized(const rmw_context_t * c)
```
adamsj-ros: These are pretty short functions so it's less important to use meaningful variable names, but I feel like it's always better practice to use a variable name that shows intent rather than a single character. It's also harder to search a single character variable name in a broader context like a workspace.

adamsj-ros (changes requested):

rmw_opendds_cpp/src/rmw_init.cpp
lij-oci: That was a temporary number for testing. It has been put back to 0.

rmw_opendds_cpp/src/rmw_init.cpp
lij-oci: It's checked in is_zero_initialized(dst).

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/RmwAllocateFree.hpp
lij-oci: Just to be consistent with rmw code.

rmw_opendds_shared_cpp/src/init.cpp
lij-oci: done

rmw_opendds_shared_cpp/src/init.cpp
lij-oci: done

rmw_opendds_shared_cpp/src/init.cpp
lij-oci: done

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/RmwAllocateFree.hpp
lij-oci: This template class RmwAllocateFree is used to define custom class specific creation and destruction using rmw allocate/free. The parameter list is checked at compiling time to match the custom class constructor. No mismatching is allowed as in free functions with variable args.

rmw_opendds_shared_cpp/src/init.cpp
```diff
@@ -13,17 +13,67 @@
...
+    RmwAllocateFree<rmw_context_impl_t>::destroy(context.impl);
   }
-  TheServiceParticipant->set_default_discovery(OpenDDS::DCPS::Discovery::DEFAULT_RTPS);
   return RMW_RET_OK;
 }
+
+bool is_zero_initialized(const rmw_init_options_t * opt)
+{
+  return opt && opt->instance_id == 0
+    && !(opt->implementatioptn_identifier)
```
adamsj-ros: misspelled implementatioptn_identifier

adamsj-ros (changes requested):

rmw_opendds_shared_cpp/src/init.cpp
lij-oci: done

adamsj-ros:
Tested locally. The breaking changes in the CI test result will be fixed on master. This merge is to the `system_test_cleanup` branch.

PR #56, opened 2021-01-23
Node, publisher

lij-oci:
This PR contains changes in node and publisher creation and destruction. Other changes in subscription, service and client are either temporary or incomplete, that are not in the scope of this PR review since they are made simply for compiling and testing the node and publisher at this point.

Fixes (Trello link here OR reason to fix if not on Trello)

# Type of change
Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
Inspection and system tests.

### DIFF
--- a/rmw_opendds_cpp/include/rmw_opendds_cpp/opendds_static_publisher_info.hpp
+++ b/rmw_opendds_cpp/include/rmw_opendds_cpp/opendds_static_publisher_info.hpp
@@ -17,8 +17,10 @@
 
 #include <atomic>
 
+#include "rmw_opendds_shared_cpp/OpenDDSNode.hpp"
 #include "rmw_opendds_shared_cpp/types.hpp"
 #include "rmw_opendds_shared_cpp/opendds_static_event_info.hpp"
+#include "rmw_opendds_shared_cpp/RmwAllocateFree.hpp"
 
 #include "rosidl_typesupport_opendds_cpp/message_type_support.h"
 
@@ -28,40 +30,37 @@ extern "C"
 {
 struct OpenDDSStaticPublisherInfo : OpenDDSCustomEventInfo
 {
-  DDS::Publisher_var dds_publisher_;
-  OpenDDSPublisherListener * listener_;
-  DDS::DataWriter_var topic_writer_;
   const message_type_support_callbacks_t * callbacks_;
-  rmw_gid_t publisher_gid;
-  OpenDDSStaticPublisherInfo() :
-    dds_publisher_(),
-    listener_(nullptr),
-    topic_writer_(),
-    callbacks_(nullptr),
-    publisher_gid() {}
+  const std::string topic_name_;
+  const std::string type_name_;
+  OpenDDSPublisherListener * listener_;
+  DDS::Publisher_var publisher_;
+  DDS::DataWriter_var writer_;
+  rmw_gid_t publisher_gid_;
 
-  /**
-   * Remap the OpenDDS DataWriter Status to a generic RMW status type.
-   *
-   * \param mask input status mask
-   * \param event
-   */
+  typedef RmwAllocateFree<OpenDDSStaticPublisherInfo> Raf;
+  static OpenDDSStaticPublisherInfo* get_from(const rmw_publisher_t * publisher);
+
+  // Remap the OpenDDS DataWriter Status to a generic RMW status type
   rmw_ret_t get_status(DDS::StatusMask mask, void* event) override;
 
-  /// Return the topic writer entity for this publisher.
-  /**
-   * \return the topic writer associated with this publisher
-   */
+  // return the topic writer associated with this publisher
   DDS::Entity* get_entity() override;
+private:
+  friend Raf;
+  OpenDDSStaticPublisherInfo(DDS::DomainParticipant_var dp, const rosidl_message_type_support_t& ros_ts,
+                             const char * topic_name, const rmw_qos_profile_t& rmw_qos);
+  ~OpenDDSStaticPublisherInfo(){ cleanup(); }
+  void cleanup();
 };
 }  // extern "C"
 
 class OpenDDSPublisherListener : public DDS::PublisherListener
 {
 public:
-  virtual void on_publication_matched(
-    DDS::DataWriter *,
-    const DDS::PublicationMatchedStatus & status)
+  typedef RmwAllocateFree<OpenDDSPublisherListener> Raf;
+
+  virtual void on_publication_matched(DDS::DataWriter *, const DDS::PublicationMatchedStatus & status)
   {
     current_count_ = status.current_count;
   }
@@ -71,37 +70,14 @@ public:
     return current_count_;
   }
 
-  void on_offered_deadline_missed(
-    DDS::DataWriter_ptr writer,
-    const DDS::OfferedDeadlineMissedStatus& status
-  )
-  {
-    ACE_UNUSED_ARG(writer);
-    ACE_UNUSED_ARG(status);
-    // Add your implementation here
-  }
-
-  void on_offered_incompatible_qos(
-    DDS::DataWriter_ptr writer,
-    const DDS::OfferedIncompatibleQosStatus& status
-  )
-  {
-    ACE_UNUSED_ARG(writer);
-    ACE_UNUSED_ARG(status);
-    // Add your implementation here
-  }
-
-  void on_liveliness_lost(
-    DDS::DataWriter_ptr writer,
-    const DDS::LivelinessLostStatus& status
-  )
-  {
-    ACE_UNUSED_ARG(writer);
-    ACE_UNUSED_ARG(status);
-    // Add your implementation here
-  }
+  void on_offered_deadline_missed(DDS::DataWriter_ptr, const DDS::OfferedDeadlineMissedStatus&) {}
+  void on_offered_incompatible_qos(DDS::DataWriter_ptr, const DDS::OfferedIncompatibleQosStatus&) {}
+  void on_liveliness_lost(DDS::DataWriter_ptr, const DDS::LivelinessLostStatus&) {}
 
 private:
+  friend Raf;
+  OpenDDSPublisherListener(){ current_count_ = 0; }
+  ~OpenDDSPublisherListener(){}
   std::atomic<std::size_t> current_count_;
 };
 
--- a/rmw_opendds_cpp/include/rmw_opendds_cpp/opendds_static_publisher_info.hpp
+++ b/rmw_opendds_cpp/include/rmw_opendds_cpp/opendds_static_publisher_info.hpp
@@ -50,7 +50,7 @@ private:
   friend Raf;
   OpenDDSStaticPublisherInfo(DDS::DomainParticipant_var dp, const rosidl_message_type_support_t& ros_ts,
                              const char * topic_name, const rmw_qos_profile_t& rmw_qos);
-  ~OpenDDSStaticPublisherInfo(){ cleanup(); }
+  ~OpenDDSStaticPublisherInfo() { cleanup(); }
   void cleanup();
 };
 }  // extern "C"
@@ -76,8 +76,8 @@ public:
 
 private:
   friend Raf;
-  OpenDDSPublisherListener(){ current_count_ = 0; }
-  ~OpenDDSPublisherListener(){}
+  OpenDDSPublisherListener() { current_count_ = 0; }
+  ~OpenDDSPublisherListener() {}
   std::atomic<std::size_t> current_count_;
 };
 
--- a/rmw_opendds_cpp/src/opendds_static_publisher_info.cpp
+++ b/rmw_opendds_cpp/src/opendds_static_publisher_info.cpp
@@ -22,7 +22,7 @@
 #include "rmw_opendds_shared_cpp/qos.hpp"
 #include "opendds_static_serialized_dataTypeSupportImpl.h"
 
-#include "dds/DCPS/DataWriterImpl_T.h"
+#include <dds/DCPS/DataWriterImpl_T.h>
 #include <dds/DCPS/Marked_Default_Qos.h>
 
 #include "rmw/visibility_control.h"
@@ -129,7 +129,7 @@ OpenDDSStaticPublisherInfo::OpenDDSStaticPublisherInfo(DDS::DomainParticipant_va
   , const rosidl_message_type_support_t & ros_ts
   , const char * topic_name
   , const rmw_qos_profile_t & rmw_qos
-  )
+)
   : callbacks_(static_cast<const message_type_support_callbacks_t*>(ros_ts.data))
   , topic_name_(get_topic_str(topic_name, rmw_qos.avoid_ros_namespace_conventions))
   , type_name_(callbacks_ ? _create_type_name(callbacks_) : "")
--- a/rmw_opendds_cpp/src/opendds_static_publisher_info.cpp
+++ b/rmw_opendds_cpp/src/opendds_static_publisher_info.cpp
@@ -12,12 +12,38 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "rmw_opendds_cpp/opendds_static_publisher_info.hpp"
+
+#include "rmw_opendds_shared_cpp/identifier.hpp"
+#include "rmw_opendds_shared_cpp/OpenDDSNode.hpp"
+#include "process_topic_and_service_names.hpp"
+#include "type_support_common.hpp"
+#include "rmw_opendds_shared_cpp/event_converter.hpp"
+#include "rmw_opendds_shared_cpp/qos.hpp"
+#include "opendds_static_serialized_dataTypeSupportImpl.h"
+
+#include "dds/DCPS/DataWriterImpl_T.h"
+#include <dds/DCPS/Marked_Default_Qos.h>
+
 #include "rmw/visibility_control.h"
 #include "rmw/incompatible_qos_events_statuses.h"
 
-#include "rmw_opendds_cpp/opendds_static_publisher_info.hpp"
-#include "rmw_opendds_shared_cpp/event_converter.hpp"
-#include "rmw_opendds_shared_cpp/qos.hpp"
+OpenDDSStaticPublisherInfo * OpenDDSStaticPublisherInfo::get_from(const rmw_publisher_t * publisher)
+{
+  if (!publisher) {
+    RMW_SET_ERROR_MSG("publisher is null");
+    return nullptr;
+  }
+  if (!check_impl_id(publisher->implementation_identifier)) {
+    return nullptr;
+  }
+  auto publisher_info = static_cast<OpenDDSStaticPublisherInfo *>(publisher->data);
+  if (!publisher_info) {
+    RMW_SET_ERROR_MSG("publisher_info is null");
+    return nullptr;
+  }
+  return publisher_info;
+}
 
 rmw_ret_t OpenDDSStaticPublisherInfo::get_status(
   DDS::StatusMask mask,
@@ -27,9 +53,7 @@ rmw_ret_t OpenDDSStaticPublisherInfo::get_status(
     case DDS::LIVELINESS_LOST_STATUS:
       {
         DDS::LivelinessLostStatus liveliness_lost;
-        DDS::ReturnCode_t dds_return_code =
-          topic_writer_->get_liveliness_lost_status(liveliness_lost);
-
+        DDS::ReturnCode_t dds_return_code = writer_->get_liveliness_lost_status(liveliness_lost);
         rmw_ret_t from_dds = check_dds_ret_code(dds_return_code);
         if (from_dds != RMW_RET_OK) {
           return from_dds;
@@ -45,9 +69,7 @@ rmw_ret_t OpenDDSStaticPublisherInfo::get_status(
     case DDS::OFFERED_DEADLINE_MISSED_STATUS:
       {
         DDS::OfferedDeadlineMissedStatus offered_deadline_missed;
-        DDS::ReturnCode_t dds_return_code =
-          topic_writer_->get_offered_deadline_missed_status(offered_deadline_missed);
-
+        DDS::ReturnCode_t dds_return_code = writer_->get_offered_deadline_missed_status(offered_deadline_missed);
         rmw_ret_t from_dds = check_dds_ret_code(dds_return_code);
         if (from_dds != RMW_RET_OK) {
           return from_dds;
@@ -64,9 +86,7 @@ rmw_ret_t OpenDDSStaticPublisherInfo::get_status(
     case DDS::OFFERED_INCOMPATIBLE_QOS_STATUS:
       {
         DDS::OfferedIncompatibleQosStatus offered_incompatible_qos;
-        DDS::ReturnCode_t dds_return_code =
-          topic_writer_->get_offered_incompatible_qos_status(offered_incompatible_qos);
-
+        DDS::ReturnCode_t dds_return_code = writer_->get_offered_incompatible_qos_status(offered_incompatible_qos);
         rmw_ret_t from_dds = check_dds_ret_code(dds_return_code);
         if (RMW_RET_OK != from_dds) {
           return from_dds;
@@ -90,5 +110,90 @@ rmw_ret_t OpenDDSStaticPublisherInfo::get_status(
 
 DDS::Entity * OpenDDSStaticPublisherInfo::get_entity()
 {
-  return topic_writer_;
+  return writer_;
+}
+
+void OpenDDSStaticPublisherInfo::cleanup()
+{
+  if (writer_) {
+    writer_ = nullptr;
+  }
+  if (publisher_) {
+    publisher_ = nullptr;
+  }
+  OpenDDSPublisherListener::Raf::destroy(listener_);
+  callbacks_ = nullptr;
+}
+
+OpenDDSStaticPublisherInfo::OpenDDSStaticPublisherInfo(DDS::DomainParticipant_var dp
+  , const rosidl_message_type_support_t & ros_ts
+  , const char * topic_name
+  , const rmw_qos_profile_t & rmw_qos
+  )
+  : callbacks_(static_cast<const message_type_support_callbacks_t*>(ros_ts.data))
+  , topic_name_(get_topic_str(topic_name, rmw_qos.avoid_ros_namespace_conventions))
+  , type_name_(callbacks_ ? _create_type_name(callbacks_) : "")
+  , listener_(OpenDDSPublisherListener::Raf::create())
+  , publisher_()
+  , writer_()
+  , publisher_gid_{opendds_identifier, {0}}
+{
+  try {
+    if (!callbacks_) {
+      throw std::runtime_error("callbacks_ is null");
+    }
+    if (topic_name_.empty()) {
+      throw std::runtime_error("topic_name_ is empty");
+    }
+    if (type_name_.empty()) {
+      throw std::runtime_error("type_name_ is empty");
+    }
+    if (!listener_) {
+      throw std::runtime_error("OpenDDSPublisherListener failed");
+    }
+    if (!dp) {
+      throw std::runtime_error("DomainParticipant is null");
+    }
+    publisher_ = dp->create_publisher(PUBLISHER_QOS_DEFAULT, listener_, DDS::PUBLICATION_MATCHED_STATUS);
+    if (!publisher_) {
+      throw std::runtime_error("create_publisher failed");
+    }
+    OpenDDSStaticSerializedDataTypeSupport_var ts = new OpenDDSStaticSerializedDataTypeSupportImpl();
+    if (ts->register_type(dp, type_name_.c_str()) != DDS::RETCODE_OK) {
+      throw std::runtime_error("failed to register OpenDDS type");
+    }
+
+    // find or create DDS::Topic
+    DDS::TopicDescription_var td = dp->lookup_topicdescription(topic_name_.c_str());
+    DDS::Topic_var topic = td ? dp->find_topic(topic_name_.c_str(), DDS::Duration_t{0, 0}) :
+      dp->create_topic(topic_name_.c_str(), type_name_.c_str(), TOPIC_QOS_DEFAULT, NULL, OpenDDS::DCPS::NO_STATUS_MASK);
+    if (!topic) {
+      throw std::runtime_error(std::string("failed to ") + (td ? "find" : "create") + " topic");
+    }
+
+    DDS::DataWriterQos dw_qos;
+    if (!get_datawriter_qos(publisher_.in(), rmw_qos, dw_qos)) {
+      throw std::runtime_error("get_datawriter_qos failed");
+    }
+    writer_ = publisher_->create_datawriter(topic, dw_qos, NULL, OpenDDS::DCPS::NO_STATUS_MASK);
+    if (!writer_) {
+      throw std::runtime_error("create_datawriter failed");
+    }
+
+    auto wri = dynamic_cast<OpenDDS::DCPS::DataWriterImpl_T<OpenDDSStaticSerializedData>*>(writer_.in());
+    wri->set_marshal_skip_serialize(true);
+
+    static_assert(sizeof(OpenDDSPublisherGID) <= RMW_GID_STORAGE_SIZE, "insufficient RMW_GID_STORAGE_SIZE");
+    auto gid = reinterpret_cast<OpenDDSPublisherGID*>(publisher_gid_.data);
+    gid->publication_handle = writer_->get_instance_handle();
+
+  } catch (const std::exception& e) {
+    RMW_SET_ERROR_MSG(e.what());
+    cleanup();
+    throw;
+  } catch (...) {
+    RMW_SET_ERROR_MSG("OpenDDSStaticPublisherInfo construtor failed");
+    cleanup();
+    throw;
+  }
 }
--- a/rmw_opendds_cpp/src/opendds_static_publisher_info.cpp
+++ b/rmw_opendds_cpp/src/opendds_static_publisher_info.cpp
@@ -149,7 +149,7 @@ OpenDDSStaticPublisherInfo::OpenDDSStaticPublisherInfo(DDS::DomainParticipant_va
       throw std::runtime_error("type_name_ is empty");
     }
     if (!listener_) {
-      throw std::runtime_error("OpenDDSPublisherListener failed");
+      throw std::runtime_error("OpenDDSPublisherListener failed to contstruct");
     }
     if (!dp) {
       throw std::runtime_error("DomainParticipant is null");
--- a/rmw_opendds_cpp/src/rmw_node.cpp
+++ b/rmw_opendds_cpp/src/rmw_node.cpp
@@ -17,8 +17,8 @@
 #include "rmw/rmw.h"
 
 #include "rmw_opendds_shared_cpp/node.hpp"
-
-#include "rmw_opendds_cpp/identifier.hpp"
+#include "rmw_opendds_shared_cpp/identifier.hpp"
+#include "rmw_opendds_shared_cpp/OpenDDSNode.hpp"
 
 extern "C"
 {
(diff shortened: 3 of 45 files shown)


rmw_opendds_cpp/include/rmw_opendds_cpp/opendds_static_publisher_info.hpp
```diff
@@ -28,40 +30,37 @@ extern "C"
...
-  /**
-   * \return the topic writer associated with this publisher
-   */
+  // return the topic writer associated with this publisher
   DDS::Entity* get_entity() override;
+private:
+  friend Raf;
+  OpenDDSStaticPublisherInfo(DDS::DomainParticipant_var dp, const rosidl_message_type_support_t& ros_ts,
+                             const char * topic_name, const rmw_qos_profile_t& rmw_qos);
+  ~OpenDDSStaticPublisherInfo(){ cleanup(); }
```
sonndinh:
```suggestion
  ~OpenDDSStaticPublisherInfo() { cleanup(); }
```

rmw_opendds_cpp/include/rmw_opendds_cpp/opendds_static_publisher_info.hpp
```diff
@@ -71,37 +70,14 @@ class OpenDDSPublisherListener : public DDS::PublisherListener
...
-    ACE_UNUSED_ARG(status);
-    // Add your implementation here
-  }
+  void on_offered_deadline_missed(DDS::DataWriter_ptr, const DDS::OfferedDeadlineMissedStatus&) {}
+  void on_offered_incompatible_qos(DDS::DataWriter_ptr, const DDS::OfferedIncompatibleQosStatus&) {}
+  void on_liveliness_lost(DDS::DataWriter_ptr, const DDS::LivelinessLostStatus&) {}
 
 private:
+  friend Raf;
+  OpenDDSPublisherListener(){ current_count_ = 0; }
```
sonndinh:
```suggestion
  OpenDDSPublisherListener() { current_count_ = 0; }
```

rmw_opendds_cpp/include/rmw_opendds_cpp/opendds_static_publisher_info.hpp
```diff
@@ -71,37 +70,14 @@ class OpenDDSPublisherListener : public DDS::PublisherListener
...
+  ~OpenDDSPublisherListener(){}
```
sonndinh:
```suggestion
  ~OpenDDSPublisherListener() {}
```

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/OpenDDSNode.hpp
```diff
@@ -0,0 +1,57 @@
+// Copyright 2015 Open Source Robotics Foundation, Inc.
```
sonndinh, line 1: Maybe change the file name to `opendds_node.hpp` so that it's consistent with all other files?

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/OpenDDSNode.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include <dds/DCPS/DomainParticipantImpl.h>
+
+class OpenDDSNode
+{
+public:
+  typedef RmwAllocateFree<OpenDDSNode> Raf;
+  static OpenDDSNode* get_from(const rmw_node_t * node);
+  const rmw_guard_condition_t * get_guard_condition() const { return gc_; }
+  bool assert_liveliness() const { return dp_->assert_liveliness() != DDS::RETCODE_OK; }
```
sonndinh: Do you mean this?
```
  bool assert_liveliness() const { return dp_->assert_liveliness() == DDS::RETCODE_OK; }
```

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/OpenDDSNode.hpp
```diff
@@ -0,0 +1,57 @@
...
+  void add_sub(const DDS::InstanceHandle_t& sub, const std::string& topic_name, const std::string& type_name);
+  bool remove_pub(const DDS::InstanceHandle_t& pub);
+  bool remove_sub(const DDS::InstanceHandle_t& sub);
+  rmw_ret_t count_publishers(const char * topic_name, size_t * count);
+  rmw_ret_t count_subscribers(const char * topic_name, size_t * count);
+
+private:
+  friend Raf;
+  OpenDDSNode(rmw_context_t & context);
+  ~OpenDDSNode(){ cleanup(); }
```
sonndinh:
```suggestion
  ~OpenDDSNode() { cleanup(); }
```

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/OpenDDSNode.hpp
```diff
@@ -0,0 +1,57 @@
...
+  typedef RmwAllocateFree<OpenDDSNode> Raf;
+  static OpenDDSNode* get_from(const rmw_node_t * node);
+  const rmw_guard_condition_t * get_guard_condition() const { return gc_; }
+  bool assert_liveliness() const { return dp_->assert_liveliness() != DDS::RETCODE_OK; }
+  DDS::DomainParticipant_var dp() { return dp_; }
+  void add_pub(const DDS::InstanceHandle_t& pub, const std::string& topic_name, const std::string& type_name);
+  void add_sub(const DDS::InstanceHandle_t& sub, const std::string& topic_name, const std::string& type_name);
+  bool remove_pub(const DDS::InstanceHandle_t& pub);
+  bool remove_sub(const DDS::InstanceHandle_t& sub);
+  rmw_ret_t count_publishers(const char * topic_name, size_t * count);
```
sonndinh: I don't remember the convention of placing the reference `&` and pointer `*` signs in this repository but looks like they're not using consistently throughout this file and other files -- sometimes they're placed right next to the type and sometimes there are spaces in between.

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/node.hpp
```diff
@@ -15,29 +15,16 @@
...
-RMW_OPENDDS_SHARED_CPP_PUBLIC
-rmw_ret_t
-destroy_node(const char * implementation_identifier, rmw_node_t * node);
+#include "rmw/types.h"
 
 RMW_OPENDDS_SHARED_CPP_PUBLIC
-const rmw_guard_condition_t *
-node_get_graph_guard_condition(const rmw_node_t * node);
+rmw_node_t *
+create_node(rmw_context_t & context, const char * name, const char * namespace_);
```
sonndinh:
```suggestion
create_node(rmw_context_t & context, const char * name, const char * namespace);
```

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/types.hpp
```diff
@@ -15,6 +15,15 @@
 #ifndef RMW_OPENDDS_SHARED_CPP__TYPES_HPP_
 #define RMW_OPENDDS_SHARED_CPP__TYPES_HPP_
 
+#include "topic_cache.hpp"
+#include "rmw_opendds_shared_cpp/RmwAllocateFree.hpp"
+#include "rmw_opendds_shared_cpp/opendds_include.hpp"
+#include "rmw_opendds_shared_cpp/visibility_control.h"
```
sonndinh, line 21:
```suggestion
#include "RmwAllocateFree.hpp"
#include "opendds_include.hpp"
#include "visibility_control.h"
```

rmw_opendds_shared_cpp/src/OpenDDSNode.cpp
```diff
@@ -0,0 +1,241 @@
+// Copyright 2015-2017 Open Source Robotics Foundation, Inc.
```
sonndinh, line 1: Maybe rename the file to `opendds_node.cpp`?

rmw_opendds_shared_cpp/src/OpenDDSNode.cpp
```diff
@@ -0,0 +1,241 @@
...
+#include "rmw_opendds_shared_cpp/init.hpp"
+#include "rmw_opendds_shared_cpp/node.hpp"
+#include "rmw_opendds_shared_cpp/types.hpp"
+#include "rmw_opendds_shared_cpp/identifier.hpp"
+
+#include "rmw/allocators.h"
+#include "rmw/error_handling.h"
+#include "rmw/impl/cpp/macros.hpp"
+
+#include "dds/DdsDcpsCoreTypeSupportC.h"
```
sonndinh:
```suggestion
#include <dds/DdsDcpsCoreTypeSupportC.h>
```

rmw_opendds_shared_cpp/src/OpenDDSNode.cpp
```diff
@@ -0,0 +1,241 @@
...
+#include <dds/DCPS/Marked_Default_Qos.h>
+#include "dds/DCPS/Service_Participant.h"
+#include "dds/DCPS/BuiltInTopicUtils.h"
```
sonndinh:
```suggestion
#include <dds/DCPS/Service_Participant.h>
#include <dds/DCPS/BuiltInTopicUtils.h>
```

rmw_opendds_shared_cpp/src/OpenDDSNode.cpp
```diff
@@ -0,0 +1,241 @@
...
+      throw std::runtime_error("builtin subscription datareader is null");
+    }
+    sub_dr->set_listener(sub_listener_, DDS::DATA_AVAILABLE_STATUS);
+
+  } catch (const std::exception& e) {
+    RMW_SET_ERROR_MSG(e.what());
+    cleanup();
+    throw;
+  } catch (...) {
+    RMW_SET_ERROR_MSG("OpenDDSNode construtor failed");
```
sonndinh:
```suggestion
    RMW_SET_ERROR_MSG("OpenDDSNode constructor failed");
```

rmw_opendds_shared_cpp/src/node.cpp
```diff
@@ -12,19 +12,20 @@
...
 
 #include "rmw/allocators.h"
 #include "rmw/error_handling.h"
 #include "rmw/impl/cpp/macros.hpp"
 
+#include "rcutils/filesystem.h"
+
 #include "dds/DdsDcpsCoreTypeSupportC.h"
 #include "dds/DCPS/Service_Participant.h"
 #include "dds/DCPS/BuiltInTopicUtils.h"
```
sonndinh, line 31:
```suggestion
#include <dds/DdsDcpsCoreTypeSupportC.h>
#include <dds/DCPS/Service_Participant.h>
#include <dds/DCPS/BuiltInTopicUtils.h>
```

rmw_opendds_cpp/src/opendds_static_publisher_info.cpp
```diff
@@ -12,12 +12,38 @@
...
+
+#include "rmw_opendds_shared_cpp/identifier.hpp"
+#include "rmw_opendds_shared_cpp/OpenDDSNode.hpp"
+#include "process_topic_and_service_names.hpp"
+#include "type_support_common.hpp"
 #include "rmw_opendds_shared_cpp/event_converter.hpp"
 #include "rmw_opendds_shared_cpp/qos.hpp"
+#include "opendds_static_serialized_dataTypeSupportImpl.h"
+
+#include "dds/DCPS/DataWriterImpl_T.h"
```
sonndinh:
```suggestion
#include <dds/DCPS/DataWriterImpl_T.h>
```

rmw_opendds_cpp/src/opendds_static_publisher_info.cpp
```diff
@@ -90,5 +110,90 @@ rmw_ret_t OpenDDSStaticPublisherInfo::get_status(
...
+  }
+  OpenDDSPublisherListener::Raf::destroy(listener_);
+  callbacks_ = nullptr;
+}
+
+OpenDDSStaticPublisherInfo::OpenDDSStaticPublisherInfo(DDS::DomainParticipant_var dp
+  , const rosidl_message_type_support_t & ros_ts
+  , const char * topic_name
+  , const rmw_qos_profile_t & rmw_qos
+  )
```
sonndinh:
```suggestion
  , const rmw_qos_profile_t & rmw_qos)
```

rmw_opendds_cpp/src/rmw_publisher.cpp
```diff
@@ -22,15 +22,12 @@
...
+#include "rmw_opendds_shared_cpp/identifier.hpp"
 
 #include "process_topic_and_service_names.hpp"
 #include "type_support_common.hpp"
 #include "rmw_opendds_cpp/opendds_static_publisher_info.hpp"
 
-// include patched generated code from the build folder
-//#include "opendds_static_serialized_dataTypeSupportC.h"
 #include "opendds_static_serialized_dataTypeSupportImpl.h"
 #include "dds/DCPS/DataWriterImpl_T.h"
```
sonndinh:
```suggestion
#include <dds/DCPS/DataWriterImpl_T.h>
```

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/OpenDDSNode.hpp
lij-oci: Good catch! Thanks.

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/node.hpp
lij-oci: I don't like namespace_ as a parameter either, since underscore suffix is usually used for member variables. So I changed it to name_space in the .cpp file, but for got it here in .hpp. I understand the original namespace_ is to avoid collision with the cpp keyword namespace.

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/OpenDDSNode.hpp
lij-oci: Since our OpenDDS standard is that they attach to the type, I habitually do that. Last year we found most code in our rmw repos put them in the middle, so we started to follow suit. We will leave it eventually to the rmw linters or beautifiers.

rmw_opendds_cpp/src/opendds_static_publisher_info.cpp
lij-oci: I think the purpose of putting a comma in front of a line is to avoid changes to  unaffected lines when one or more lines are later changed (added, removed or reordered). In this case rmw_qos is the last parameter now. If we add one later, the rmw_qos line stays the same in git repo. Moving ')' to the end of rmw_qos cancels the purpose of a front comma. With that said, I removed the 2 spaces before ')' so it can be a clear separator between the parameters and the initializations.

rmw_opendds_cpp/src/rmw_subscription.cpp
```diff
@@ -74,6 +74,7 @@ clean_subscription(rmw_subscription_t * subscription, OpenDDSNodeInfo & node_inf
 
   auto info = static_cast<OpenDDSStaticSubscriberInfo*>(subscription->data);
   if (info) {
+/*
```
adamsj-ros: What's the TODO here? Otherwise I recommend removing this block.

rmw_opendds_cpp/src/rmw_subscription.cpp
```diff
@@ -280,7 +276,7 @@ rmw_create_subscription(
   } catch (...) {
     RMW_SET_ERROR_MSG("rmw_create_subscription failed");
   }
-  clean_subscription(subscription, *node_info, *node_info->participant);
+//  clean_subscription(subscription, *node_info, *node_info->participant);
```
adamsj-ros, line 280: TODO or remove commented code.

rmw_opendds_cpp/include/rmw_opendds_cpp/opendds_static_publisher_info.hpp
adamsj-ros: Why wouldn't the contents of `cleanup()` be in the destructor? Just started my review so maybe I'll see it later.

rmw_opendds_cpp/src/opendds_static_publisher_info.cpp
```diff
@@ -90,5 +110,90 @@ rmw_ret_t OpenDDSStaticPublisherInfo::get_status(
...
+      throw std::runtime_error("callbacks_ is null");
+    }
+    if (topic_name_.empty()) {
+      throw std::runtime_error("topic_name_ is empty");
+    }
+    if (type_name_.empty()) {
+      throw std::runtime_error("type_name_ is empty");
+    }
+    if (!listener_) {
+      throw std::runtime_error("OpenDDSPublisherListener failed");
```
adamsj-ros: Maybe change to `"OpenDDSPublisherListener failed to contstruct"`?

rmw_opendds_cpp/src/rmw_node.cpp
```diff
@@ -30,47 +30,39 @@ rmw_create_node(
   size_t domain_id,
   bool localhost_only)
 {
-  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context, NULL);
-  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
-    init context,
-    context->implementation_identifier,
-    opendds_identifier,
-    // TODO(wjwwood): replace this with RMW_RET_INCORRECT_RMW_IMPLEMENTATION when refactored
```
adamsj-ros, line 38: Should we leave this TODO in?

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/types.hpp
lij-oci: For the following reasons, I propose not to make these changes in this PR:

1) Currently the rmw repos use double quotes to include most files except standard library files. This definitely differs from our OpenDDS standard regarding double quotes and angle brackets. The order of includes is even worse. Sometimes when I made substantial changes in some files, I also changed the include quotes and order, but not systematically change all of them yet.

2) The rmw src files are organized in such directories:
rmw_opendds_cpp/include/rmw_opendds_cpp/identifier.hpp
rmw_opendds_cpp/src/identifier.cpp

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/init.hpp
rmw_opendds_shared_cpp/src/init.cpp

For a cpp file, the same name header file should be first included (in double quotes if in the same directory). However, in rmw_opendds_shared_cpp/src/init.cpp, for example, we have #include "rmw_opendds_shared_cpp/init.hpp". Since they are not in the same directories, should we go ahead replace all of such includes with angle brackets? 

It is not at all unlikely that different rmw implementations have the same filenames, such as identifier.hpp or init.hpp. Do we feel safe to drop the directories ("rmw_opendds_cpp/" or "rmw_opendds_shared_cpp/") from the includes now when we are right in the middle of trying to figure out some problems including why we did not get our own implementation_identifier? At least the directories have _opendds_ in it. It's best that we focus on the intended changes here and hold the other things later.

So I'll create a jira card for investigating our rmw includes. (Maybe rmw already has a linter for it.)

https://opendds.atlassian.net/jira/software/projects/RMW/boards/12?selectedIssue=RMW-91

rmw_opendds_cpp/src/rmw_node.cpp
```diff
@@ -30,47 +30,39 @@ rmw_create_node(
...
-    return NULL);
-  return create_node(opendds_identifier, name, namespace_, domain_id);
+  (void)domain_id;
+  (void)localhost_only;
+  RMW_CHECK_FOR_NULL_WITH_MSG(context, "context is null", NULL);
+  if (!check_impl_id(context->implementation_identifier)) {
+    return NULL;
```
adamsj-ros, line 37: line 35 prints an error. Shouldn't this also print an error?

rmw_opendds_cpp/src/rmw_node.cpp
```diff
@@ -30,47 +30,39 @@ rmw_create_node(
...
+  return create_node(*context, name, namespace_);
 }
 
 rmw_ret_t
 rmw_destroy_node(rmw_node_t * node)
 {
-  return destroy_node(opendds_identifier, node);
+  RMW_CHECK_FOR_NULL_WITH_MSG(node, "node is null", RMW_RET_INVALID_ARGUMENT);
+  if (!check_impl_id(node->implementation_identifier)) {
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION;
```
adamsj-ros, line 50: Shouldn't this print an error here? Maybe it's handled by the caller but since we already have these other null checks printing, its seem well established to print errors at this level.

rmw_opendds_cpp/src/rmw_node.cpp
```diff
@@ -30,47 +30,39 @@ rmw_create_node(
...
-  RMW_CHECK_ARGUMENT_FOR_NULL(node, RMW_RET_INVALID_ARGUMENT);
-  RMW_CHECK_ARGUMENT_FOR_NULL(node->implementation_identifier, RMW_RET_INVALID_ARGUMENT);
-  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
-    node handle,
-    node->implementation_identifier,
-    opendds_identifier,
-    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION)
-  return node_assert_liveliness(node);
+  auto dds_node = OpenDDSNode::get_from(node);
+  return dds_node && dds_node->assert_liveliness() ? RMW_RET_OK : RMW_RET_ERROR;
```
adamsj-ros, line 59: print error?

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/OpenDDSNode.hpp
lij-oci: The class OpenDDSNode, defined in OpenDDSNode.hpp and implemented in OpenDDSNode.cpp, was originally defined as a c-struct: OpenDDSNodeInfo in a c-style header file types.hpp (which contains several classes and structs). The functions related to OpenDDSNodeInfo were defined in multip[le c function files.

My first implementation of OpenDDSNode was made in the existing files. As more funcionalities were developed, the code became messy, and the class was moved into its own .hpp/.cpp files. At first I struggled about the file naming, and then decided to follow the OpenDDS (also generally accepted) c++ file naming convention, for these reasons:

1) To name them as opendds_node.hpp/.cpp is following the c convention, and I would have to name the files of the class RmwAllocateFree as rmw_allocate_free.hpp/.cpp, that would easily be confused with rmw function files instead of the c++ class files.

2) These classes are not exposed to applications using rmw, i.e., not RMW_OPENDDS_SHARED_CPP_PUBLIC prefixed. They are just OpenDDS implementations of some rmw function interfaces. The only difference is now we choose to use c++ class to encapsulate these implementations, instead of using the c-style types.hpp and several c function files. 

When we complete the changes for all the objects using OpenDDSNode, the c-struct OpenDDSNodeInfo and some c functions files will be removed.

rmw_opendds_cpp/src/rmw_service.cpp
```diff
@@ -40,6 +40,7 @@ extern "C"
 rmw_ret_t clean_service(OpenDDSNodeInfo & node_info, rmw_service_t * service)
 {
   auto ret = RMW_RET_OK;
+  return ret; //?? temp
```
adamsj-ros, line 43: What's the comment mean?

rmw_opendds_cpp/src/rmw_service.cpp
```diff
@@ -97,30 +98,24 @@ rmw_create_service(
...
-
-  auto node_info = static_cast<OpenDDSNodeInfo *>(node->data);
-  RMW_CHECK_FOR_NULL_WITH_MSG(node_info, "node_info is null", return nullptr);
-  RMW_CHECK_FOR_NULL_WITH_MSG(node_info->participant.in(), "participant is null", return nullptr);
-  RMW_CHECK_FOR_NULL_WITH_MSG(node_info->publisher_listener, "publisher_listener is null", return nullptr);
-  RMW_CHECK_FOR_NULL_WITH_MSG(node_info->subscriber_listener, "subscriber_listener is null", return nullptr);
-
+  auto dds_node = OpenDDSNode::get_from(node);
+  if (!dds_node) {
+    return nullptr; // error set in get_from
```
adamsj-ros, line 103: print error? I think most people would consider returning nullptr as an error but it could also be considered normal behavior. I think in this case since it's an error we want to print at this level.

rmw_opendds_shared_cpp/src/OpenDDSNode.cpp
lij-oci: See comment in opendds_node.cpp.

rmw_opendds_cpp/src/rmw_service.cpp
```diff
@@ -209,7 +203,7 @@ rmw_create_service(
     if (!name || !type_name) {
       throw std::string("topicdescription name or type_name is null");
     }
-
+/*
```
adamsj-ros: TODO or remove commented code?

rmw_opendds_cpp/src/rmw_service.cpp
```diff
@@ -237,18 +231,20 @@ rmw_create_service(
...
-
+*/
     return service;
 
   } catch (const std::string& e) {
     RMW_SET_ERROR_MSG(e.c_str());
   } catch (...) {
     RMW_SET_ERROR_MSG("rmw_create_service failed");
   }
+/*
```
adamsj-ros: TODO or remove?

rmw_opendds_cpp/src/rmw_subscription.cpp
```diff
@@ -132,13 +134,15 @@ rmw_create_subscription(
...
-  RMW_CHECK_FOR_NULL_WITH_MSG(node, "node is null", return nullptr);
-  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(node, node->implementation_identifier, opendds_identifier, return nullptr)
-
-  auto node_info = static_cast<OpenDDSNodeInfo *>(node->data);
-  RMW_CHECK_FOR_NULL_WITH_MSG(node_info, "node_info is null", return nullptr);
-  RMW_CHECK_FOR_NULL_WITH_MSG(node_info->participant.in(), "participant is null", return nullptr);
-  RMW_CHECK_FOR_NULL_WITH_MSG(node_info->subscriber_listener, "subscriber_listener is null", return nullptr);
+  auto dds_node = OpenDDSNode::get_from(node);
+  if (!dds_node) {
+    return nullptr; // error set in get_from
```
adamsj-ros, line 139: print error?

rmw_opendds_shared_cpp/src/OpenDDSNode.cpp
```diff
@@ -0,0 +1,241 @@
...
+#ifdef OPENDDS_SECURITY
+#include <dds/DCPS/security/framework/Properties.h>
+#endif
+#include <string>
+
+OpenDDSNode* OpenDDSNode::get_from(const rmw_node_t * node)
+{
+  RMW_CHECK_FOR_NULL_WITH_MSG(node, "node is null", return nullptr);
+  if (!check_impl_id(node->implementation_identifier)) {
+    return nullptr;
```
adamsj-ros, line 47: print error?

rmw_opendds_shared_cpp/src/OpenDDSNode.cpp
```diff
@@ -0,0 +1,241 @@
...
+    }
+    dp_ = context.impl->dpf_->create_participant(
+      static_cast<DDS::DomainId_t>(context.options.domain_id), PARTICIPANT_QOS_DEFAULT, 0, 0);
+    if (!dp_) {
+      throw std::runtime_error("create_participant failed");
+    }
+    dpi_ = dynamic_cast<OpenDDS::DCPS::DomainParticipantImpl*>(dp_.in());
+    if (!dpi_) {
+      throw std::runtime_error("casting to DomainParticipantImpl failed");
+    }
```
adamsj-ros, line 151: Thought I saw a large portion of this is duplicated from earlier. Might be good to have a helper function to improve maintenance of duplicated blocks. Then again it could be my mistake.

adamsj-ros (changes requested):
Looks good. I see what appears to be inconsistent handling (or announcement) of errors. Might want to check `nullptr` returns and print something since in a few more situations.

rmw_opendds_cpp/src/rmw_subscription.cpp
lij-oci: added comment:
//temporarily commented out for testing OpenDDSNode and publisher

rmw_opendds_cpp/include/rmw_opendds_cpp/opendds_static_publisher_info.hpp
lij-oci: Because the constructor may need to call it too.

rmw_opendds_cpp/src/rmw_node.cpp
lij-oci: I don't think it's necessary, since when the interface changes to return the object in an out parameter, the function return code here is obvious.

rmw_opendds_cpp/src/rmw_subscription.cpp
lij-oci: done in get_from().

rmw_opendds_shared_cpp/src/OpenDDSNode.cpp
lij-oci: done in check_impl_id().

rmw_opendds_shared_cpp/src/OpenDDSNode.cpp
lij-oci: Maybe you saw the same logic in constructing OpenDDSStaticPublisherInfo, but they are creating different things, no duplicate. One thing I could simplify is to take out context_.implementation_identifier. Since that will affect three classes and we want this PR to be merged first, I decided to do that in next PR.

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/node.hpp
sonndinh: Yeah, `name_space` is better, I forgot `namespace` is a keyword.

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/types.hpp
sonndinh: > For a cpp file, the same name header file should be first included (in double quotes if in the same directory). However, in rmw_opendds_shared_cpp/src/init.cpp, for example, we have #include "rmw_opendds_shared_cpp/init.hpp". Since they are not in the same directories, should we go ahead replace all of such includes with angle brackets?

I think using double quotes include in a cpp in the same repo file is ok since the included file like "rmw_opendds_shared_cpp/init.hpp" is in a relative path to the source file. This is similar to the include rules of OpenDDS.

> It is not at all unlikely that different rmw implementations have the same filenames, such as identifier.hpp or init.hpp. Do we feel safe to drop the directories ("rmw_opendds_cpp/" or "rmw_opendds_shared_cpp/") from the includes now when we are right in the middle of trying to figure out some problems including why we did not get our own implementation_identifier? At least the directories have opendds in it. It's best that we focus on the intended changes here and hold the other things later.

These 3 header files are in the same directory as types.hpp, so I think it's safe to drop the directory path. Line 18 of types.hpp already dropped the directory path for including `topic_cache.hpp`. By the way, I can't find RmwAllocateFree.hpp in include/rmw_opendds_shared_cpp, and looks like it's not a newly created file either?

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/OpenDDSNode.hpp
sonndinh: > To name them as opendds_node.hpp/.cpp is following the c convention, and I would have to name the files of the class RmwAllocateFree as rmw_allocate_free.hpp/.cpp, that would easily be confused with rmw function files instead of the c++ class files.

I don't think the lowercase_with_underscores naming is C convention. I think it's just a convention that each repo enforces. I've seen C++ repos with different file naming conventions, Google test for example uses lowercase-dash naming (https://github.com/google/googletest/tree/master/googletest/src).

sonndinh (commented):
I don't have further comments other than the existing ones. Once they are resolved, the PR is good to go.

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/OpenDDSNode.hpp
lij-oci: The main point is not that the lowercase is c convention. The standard c++ library class files are lowercased to match their classnames, and most custom-defined c++ classes are Camel-cased and their files also match their classnames to minimize potential conflicts. That's the simplest and straight forward way, and is adopted by most c++ practices including our OpenDDS standard. In this particular case, we are defining non-exposed OpenDDS-implementation specific classes. To name them like rmw_allocate_free.hpp/.cpp has two problems: 1) the filenames do not match the classnames and they can easily be taken as the free function files; 2) the rmw interface functions usually prefix rmw_ to the function names, and their files match their function names. The include guard usually follow the filenames in some fashion. If our implementation class RmwAllocateFree is defined in rmw_allocate_free.hpp/.cpp, it just adds more confusion with rmw files. In other words, we would be trying to conform to rmw function files for no benefits (these classes are not exposed to rmw applications), but only to confuse ourselves. In this case, we will be better off just conforming to our OpenDDS standard to match the classnames, since rmw does not care whether you implement all the rmw interface functions with your own free function files, or some c++ class files. All rmw expects is that when an rmw_ function is called, there is your specific implementation to support the functionality.

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/types.hpp
lij-oci: I created a jira card for solving the include issue consistently across the board. Currently there are quite some problems in the system tests we haven't figured out yet. (We didn't run these system tests last year, so these are new found problems.) Many system tests build and run mix-matched implementations (e.g. opendds publisher and fastrtps subscriber, or cpp and py). We now have the statistics of the failed tests, and the current effort is to step by step fix the problems. If we change too many things now without clear expectations toward fixing the problems (or just because we know something can be better here or there), we may lose the consistency and comparisons with the known statistics. So unless we can now link an include statement to a system test error, I still prefer to solve the include issue consistently in another PR. The RmwAllocateFree was merged in PR#53.

lij-oci: As a matter of fact, in my previous modifications, I changed a lot of the include statements according to our OpenDDS standard. When I realized the system tests actually try to build and run mixed implementations, I put those directories back in the include statements just to avoid the risk of introducing problems without knowing. (The implementation identifier mismatch is on my mind all the time.) I want to solve the include problem once for all when we have a good state to compare to.

rmw_opendds_cpp/src/rmw_node.cpp
lij-oci: It's done in check_impl_id().

rmw_opendds_cpp/src/rmw_node.cpp
lij-oci: It's done in check_impl_id().

rmw_opendds_cpp/src/rmw_service.cpp
lij-oci: That means we will come back to change service in its own PR as noted in the Description section.

rmw_opendds_cpp/src/rmw_service.cpp
lij-oci: It's done in get_from() and the comment indicates.

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/OpenDDSNode.hpp
sonndinh: As in my previous comment, I don't think there is a widely accepted guideline on whether one should choose CamelCase, snake_case, or other types for their C++ project. It's more of a choice of the owner of each specific project. Some projects like OpenDDS or some other DDS implementations choose CamelCase, some other projects choose snake_case, e.g., Google's C++ projects seem to choose this way (https://google.github.io/styleguide/cppguide.html#File_Names). 

Also, the file name doesn't need to match the class name it contains, for example, OpenDDS's DiscoveryBase.h doesn't have a class named DiscoveryBase. The file rmw_allocate_free.hpp isn't really a problem -- we can change it so that it doesn't have the rmw prefix if we really want to. In general, I think a single repo should name its files in a consistent way.

This isn't the main point of this PR. If you want to keep the file names, I'm ok with that.

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/types.hpp
sonndinh: This is fine. I guess the includes can be cleaned up later in a separate PR.

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/OpenDDSNode.hpp
sonndinh: For reference, rmw_connext names their DDS-specific classes using CamelCase while naming the corresponding files using snake_case (for example, https://github.com/ros2/rmw_connext/blob/master/rmw_connext_cpp/src/connext_static_publisher_info.hpp).

adamsj-ros: The style has not been clearly defined for our RMW. Let's table(create a Jira issue) these issues for now that relate to style. The only thing I have said in the past is to be consistent but at this point I don't think we're going to fix all these style issues in this PR. 

Thanks.

sonndinh (approved):

adamsj-ros (approved):

PR #57, opened 2021-02-03
Subscription

lij-oci:
This PR is for fixing system test issues. It contains changes in consistent object creation and destruction, and in implementation identifier checking. Files include: rmw_subscription, DDSSubscriber, DDSPublisher, rmw_publisher, rmw_publish, and implementation identifier related files. Changes in other files, such as service and client, are either incomplete or temporary, and are not in the scope of this PR review since they are made simply for compiling and testing subscriber and publisher at this point.

## Type of change
-  Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
- system tests

# Checklist:
-  New and existing system tests pass locally with my changes (when applicable)
-  Any dependent changes have been merged and published in downstream modules


rmw_opendds_cpp/src/DDSPublisher.cpp
```diff
@@ -0,0 +1,181 @@
...
+#include "rmw_opendds_shared_cpp/identifier.hpp"
+#include "rmw_opendds_shared_cpp/qos.hpp"
+#include "rmw_opendds_shared_cpp/types.hpp"
+
+#include <dds/DCPS/DataWriterImpl_T.h>
+
+#include "rmw/visibility_control.h"
+#include "rmw/incompatible_qos_events_statuses.h"
+
+DDSPublisher * DDSPublisher::from(const rmw_publisher_t * pub)
```
adamsj-ros, line 29: Could we find something a little more meaningful than `from` for this method name?

rmw_opendds_cpp/src/DDSPublisher.cpp
```diff
@@ -0,0 +1,181 @@
+// Copyright 2014-2019 Open Source Robotics Foundation, Inc.
```
adamsj-ros, line 1: Is this entire file moved from elsewhere or is this new code?

rmw_opendds_cpp/src/DDSSubscriber.cpp
```diff
@@ -0,0 +1,179 @@
+// Copyright 2014-2019 Open Source Robotics Foundation, Inc.
```
adamsj-ros, line 1: What is the source of all this new code? Is it actually new or is essentially a rename for the file?

rmw_opendds_cpp/src/rmw_node.cpp
```diff
@@ -26,20 +26,18 @@ rmw_node_t *
 rmw_create_node(
   rmw_context_t * context,
   const char * name,
-  const char * namespace_,
-  size_t domain_id,
-  bool localhost_only)
+  const char * name_space,
+  size_t /* domain_id */,
+  bool /* localhost_only */)
```
adamsj-ros: Why are these var names commented out?

rmw_opendds_cpp/src/rmw_node.cpp
```diff
@@ -26,20 +26,18 @@ rmw_node_t *
...
   RMW_CHECK_FOR_NULL_WITH_MSG(context, "context is null", NULL);
   if (!check_impl_id(context->implementation_identifier)) {
     return NULL;
   }
   RMW_CHECK_FOR_NULL_WITH_MSG(context->impl, "context->impl is null", NULL);
   RMW_CHECK_FOR_NULL_WITH_MSG(name, "node name is null", NULL);
-  RMW_CHECK_FOR_NULL_WITH_MSG(namespace_, "node namespace_ is null", NULL);
-  return create_node(*context, name, namespace_);
+  RMW_CHECK_FOR_NULL_WITH_MSG(name_space, "node name_space is null", NULL);
+  return create_node(*context, name, name_space);
```
adamsj-ros: I think the original spelling is preferred (i.e. namespace_). That's the way it's spelled in the rest of ROS and I don't think it's a good idea to split from that if someone is searching for that in code.

rmw_opendds_cpp/src/rmw_publish.cpp
```diff
@@ -54,55 +55,33 @@ rmw_ret_t
 rmw_publish(
   const rmw_publisher_t * publisher,
   const void * ros_message,
-  rmw_publisher_allocation_t * allocation)
+  rmw_publisher_allocation_t *)
```
adamsj-ros: Why leave off the var name?

rmw_opendds_cpp/src/rmw_publish.cpp
```diff
@@ -54,55 +55,33 @@ rmw_ret_t
...
-  }
-  const message_type_support_callbacks_t * callbacks = info->callbacks_;
-  if (!callbacks) {
-    RMW_SET_ERROR_MSG("publisher info callbacks is null");
-    return RMW_RET_ERROR;
-  }
-  DDS::DataWriter_var writer = info->writer_;
-  if (!writer) {
-    RMW_SET_ERROR_MSG("topic writer is null");
-    return RMW_RET_ERROR;
```
adamsj-ros, line 84: Not all these checks are in `DDSPublisher::from`. Do we need them there?

rmw_opendds_cpp/src/rmw_publisher.cpp
```diff
@@ -102,9 +102,9 @@ rmw_create_publisher(
   const rmw_qos_profile_t * rmw_qos,
   const rmw_publisher_options_t * publisher_options)
 {
-  auto dds_node = OpenDDSNode::get_from(node);
+  auto dds_node = OpenDDSNode::from(node);
   if (!dds_node) {
-    return nullptr; // error set in get_from
+    return nullptr; // error set
```
adamsj-ros: the comment no longer makes sense. Either remove it or update it.

rmw_opendds_shared_cpp/include/rmw_opendds_shared_cpp/DDSEntity.hpp
```diff
@@ -0,0 +1,95 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
```
adamsj-ros, line 1: Is this original code or copied from another file?

rmw_opendds_cpp/src/rmw_node.cpp
lij-oci: The following is from the review of PR#56 on the same issue:
@sonndinh
Suggested change:
- create_node(rmw_context_t & context, const char * name, const char * namespace_);
+ create_node(rmw_context_t & context, const char * name, const char * namespace);
 
@lij-oci
I don't like namespace_ as a parameter either, since underscore suffix is usually used for member variables. So I changed it to name_space in the .cpp file, but forgot it here in .hpp. I understand the original namespace_ is to avoid collision with the cpp keyword namespace.
 
@sonndinh
Yeah, name_space is better, I forgot namespace is a keyword.

rmw_opendds_cpp/src/rmw_publisher.cpp
lij-oci: done

rmw_opendds_cpp/src/DDSPublisher.cpp
lij-oci: It's pretty common to use 'from' and 'to' in conversion functions, such as in the c++ standard library, std::from_chars, std::to_chars, std::to_string, etc. Here, these rmw-to-dds conversion functions are defined as static class functions (not free functions), so the meaning is very clear when they are used.
For example: 
DDSPublisher * dds_pub = DDSPublisher::from(rmw_publisher);
auto dds_node = OpenDDSNode::from(rmw_node);

rmw_opendds_cpp/src/rmw_node.cpp
lij-oci: name_space is replaced with namespace_ for consistency with rmw.h.

rmw_opendds_cpp/src/rmw_publish.cpp
lij-oci: They are checked in DDSPublisher.

rmw_opendds_cpp/include/rmw_opendds_cpp/DDSPublisher.hpp
```diff
@@ -81,4 +41,34 @@ class OpenDDSPublisherListener : public DDS::PublisherListener
...
+  rmw_ret_t to_cdr_stream(const void * ros_message, rcutils_uint8_array_t & cdr_stream);
+  DDS::DataWriter_var writer() const { return writer_; }
+
+  std::size_t matched_subscribers() const { return listener_->current_count(); }
+  DDS::InstanceHandle_t instance_handle() const { return publisher_->get_instance_handle(); }
+  rmw_gid_t gid() const { return publisher_gid_; }
+
+  // Remap the OpenDDS DataWriter status to a generic RMW status
+  rmw_ret_t get_status(const DDS::StatusMask mask, void * rmw_status) override;
+  // Return the reader associated with this publisher
```
kuznetsovmoci: It's probably "writer" in the comment

rmw_opendds_cpp/include/rmw_opendds_cpp/DDSPublisher.hpp
```diff
@@ -81,4 +41,34 @@ class OpenDDSPublisherListener : public DDS::PublisherListener
...
+  DDS::Entity * get() override { return writer_; }
```
kuznetsovmoci, line 60: Would get_writer be a better name? (same in DDSSubscriber)

rmw_opendds_cpp/include/rmw_opendds_cpp/DDSPublisher.hpp
lij-oci: Good catch. Thanks.

rmw_opendds_cpp/src/DDSSubscriber.cpp
```diff
@@ -0,0 +1,179 @@
...
+  switch (mask) {
+    case DDS::LIVELINESS_CHANGED_STATUS:
+    {
+      DDS::LivelinessChangedStatus llc;
+      rmw_ret_t ret = check_dds_ret_code(reader_->get_liveliness_changed_status(llc));
+      if (ret != RMW_RET_OK) {
+        return ret;
+      }
+      auto rmw_llc = static_cast<rmw_liveliness_changed_status_t *>(rmw_status);
+      rmw_llc->alive_count = llc.alive_count;
```
kuznetsovmoci, line 76: This section is different from the similar logic in Publisher. I am assuming it's intentional, but can't tell 100%.

rmw_opendds_cpp/include/rmw_opendds_cpp/DDSPublisher.hpp
lij-oci: This is an implementation of a pure virtual function. A writer is returned here, while a reader is returned in DDSSubscriber. (I have some thought about improving this design, but could not put in this PR, for keeping some consistency of the original code to compare the primary changes in this PR.)

rmw_opendds_cpp/src/DDSSubscriber.cpp
lij-oci: Yes, they are different between pub and sub. For example, 
one cares about OFFERED_DEADLINE_MISSED_STATUS, 
while the other REQUESTED_DEADLINE_MISSED_STATUS.

However, mistakes/typos may occur in situations like this. I tried to avoid them, but you still caught one in another place. I appreciate it. So please do the same on these two functions. Thanks.

rmw_opendds_cpp/include/rmw_opendds_cpp/DDSPublisher.hpp
kuznetsovmoci: OK, it makes sense. `get_endpoint()` might also be an option, if you like to make it more specific.

kuznetsovmoci (approved):

adamsj-ros (approved):

PR #58, opened 2021-02-16
Service related changes

lij-oci:
# Description

This PR contains service related changes for fixing system test issues, including implementations of consistent client and service creation and destruction using DDSClient, DDSServer, and Service. It also contains changes in processing rmw_node related names, namespace, topic names and types, and service names. 

## Type of change
- Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
- system tests

# Checklist:
-  New and existing system tests pass locally with my changes (when applicable)
- Any dependent changes have been merged and published in downstream modules


rmw_opendds_cpp/include/rmw_opendds_cpp/DDSSubscriber.hpp
```diff
@@ -73,14 +76,15 @@ class DDSSubscriber : public DDSEntity
...
 private:
   friend Raf;
-  DDSSubscriber(DDS::DomainParticipant_var dp, const rosidl_message_type_support_t & ros_ts,
-                const char * topic_name, const rmw_qos_profile_t & rmw_qos);
+  DDSSubscriber(DDS::DomainParticipant_var dp, const rosidl_message_type_support_t * ros_ts,
+                const char * topic_name, const rmw_qos_profile_t * rmw_qos);
   ~DDSSubscriber() { cleanup(); }
   void cleanup();
 
+  DDSTopic topic_;
```
ClaytonCalabrese, line 87: Is this used anywhere?

lij-oci: Yes, topic_ is used. See DDSSubscriber.cpp.

rmw_opendds_cpp/src/identifier.cpp
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "rmw_opendds_cpp/identifier.hpp"
+#include <rmw_opendds_cpp/identifier.hpp>
```
ClaytonCalabrese, line 15:
```suggestion
#include <rmw_opendds_cpp/identifier.hpp>

```

rmw_opendds_cpp/src/init.cpp
```diff
@@ -12,11 +12,11 @@
...
 // limitations under the License.
 
-#include "rmw_opendds_cpp/init.hpp"
-#include "rmw_opendds_cpp/RmwAllocateFree.hpp"
-#include "rmw/init_options.h"
-#include "rmw/error_handling.h"
-#include "dds/DCPS/Service_Participant.h"
+#include <rmw_opendds_cpp/init.hpp>
+#include <rmw_opendds_cpp/RmwAllocateFree.hpp>
+#include <rmw/init_options.h>
```
ClaytonCalabrese, line 18:
```suggestion
#include <rmw/init_options.h>

```

rmw_opendds_cpp/src/rmw_client.cpp
```diff
@@ -12,259 +12,158 @@
...
+#include <dds/DCPS/DomainParticipantImpl.h>
 #include <dds/DCPS/Marked_Default_Qos.h>
 
 // Uncomment this to get extra console output about discovery.
-// This affects code in this file, but there is a similar variable in:
-//   rmw_opendds_cpp/shared_functions.cpp
 // #define DISCOVERY_DEBUG_LOGGING 1
 
-#include <dds/DCPS/DomainParticipantImpl.h>
+#include <rmw/allocators.h>
```
ClaytonCalabrese, line 27: These rmw includes should be below the rmw_opendds_cpp but above the dds ones.

rmw_opendds_cpp/src/rmw_client.cpp
```diff
@@ -12,259 +12,158 @@
...
+    return RMW_RET_INVALID_ARGUMENT;
+  }
+  if (!taken) {
+    RMW_SET_ERROR_MSG("taken is null");
+    return RMW_RET_INVALID_ARGUMENT;
+  }
+  *taken = dds_client->take(request_header, ros_response);
+  return RMW_RET_OK;
 }
+
```
ClaytonCalabrese:
```suggestion
```

rmw_opendds_cpp/src/OpenDDSNode.cpp
```diff
@@ -12,33 +12,34 @@
...
 #include <dds/DCPS/transport/framework/TransportExceptions.h>
 #ifdef OPENDDS_SECURITY
 #include <dds/DCPS/security/framework/Properties.h>
 #endif
-#include <string>
+
+#include <rcutils/filesystem.h>
+#include <rcutils/strdup.h>
+
+#include <rmw/allocators.h>
```
ClaytonCalabrese, line 37: The includes in this file need reordering.

https://github.com/objectcomputing/OpenDDS/blob/master/docs/guidelines.md#includes

ClaytonCalabrese (changes requested):
Just a few minor changes.

rmw_opendds_cpp/src/rmw_client.cpp
lij-oci: Yea, we have a card created for that issue: https://opendds.atlassian.net/browse/RMW-91. For now, since most header files are not in the same directory with the cpp files, the double-quotes are replaced with angle-brackets. The order is basically from the narrowest scope to widest scope. The final rmw include rules are still needed to be defined when we come to RMW-91. (For example, when we use OpenDDS code to implement rmw functions, should OpenDDS code be considered a narrower scope? Also, for class files, the class header should be the first one included. What about the function implementation files, should we put rmw headers or our implementation headers first?)

rmw_opendds_cpp/src/rmw_client.cpp
lij-oci: The beginning empty line is also removed to match this.

rmw_opendds_cpp/src/OpenDDSNode.cpp
lij-oci: see the comment above.

ClaytonCalabrese (approved):

azure-pipelines.yml
```diff
@@ -30,7 +30,7 @@ jobs:
     displayName: filesystem check
   - script: |
       cd $(Agent.BuildDirectory)/s
-      git clone https://github.com/oci-labs/rosidl_typesupport_opendds
+      git clone --depth 1 --branch pre_xtypes https://github.com/oci-labs/rosidl_typesupport_opendds.git
```
adamsj-ros: We should revert this change.

adamsj-ros (changes requested):
Still making review but you should change this now so CI will pass.

azure-pipelines.yml
lij-oci: done

rmw_opendds_cpp/src/DDSSubscriber.cpp
```diff
@@ -122,10 +121,10 @@ void DDSSubscriber::cleanup()
 }
 
 DDSSubscriber::DDSSubscriber(DDS::DomainParticipant_var dp
-  , const rosidl_message_type_support_t & ros_ts
+  , const rosidl_message_type_support_t * ros_ts
   , const char * topic_name
-  , const rmw_qos_profile_t & rmw_qos
-) : DDSEntity(ros_ts, topic_name, rmw_qos)
+  , const rmw_qos_profile_t * rmw_qos
```
adamsj-ros, line 126: Just curious why we switched from by ref to by pointer.

rmw_opendds_cpp/src/OpenDDSNode.cpp
```diff
@@ -119,15 +127,221 @@ rmw_ret_t OpenDDSNode::count_subscribers(const char * topic_name, size_t * count
...
+    if (rcutils_string_array_init(namespaces, length, &allocator) != RCUTILS_RET_OK) {
+      throw std::runtime_error(rcutils_get_error_string().str);
+    }
+    if (enclaves) {
+      if (rcutils_string_array_init(enclaves, length, &allocator) != RCUTILS_RET_OK) {
+        throw std::runtime_error(rcutils_get_error_string().str);
+      }
+    }
+  //?? why data[0] cannot be put in the for loop?
+  //?? names->data[0] not set?
```
adamsj-ros: Are we waiting for answers here?

rmw_opendds_cpp/src/DDSSubscriber.cpp
lij-oci: to localize the null-checking and the following creation code in DDSTopic for safety and consistency for both pub and sub.

rmw_opendds_cpp/src/OpenDDSNode.cpp
lij-oci: These were the things I found when moving and simplifying the existing code. They should be in a separate PR to determine whether the processing logic here is correct or needs to be modified. It's a safer strategy to have the main framework changes verified first, then to investigate such unobvious issues.

adamsj-ros: I think it's supposed to be like this (https://github.com/ros2/rmw_connext/blob/1.0.1/rmw_connext_shared_cpp/src/node_names.cpp#L111-L131) since this is the code it was created from.

lij-oci: added node names->data[0]

adamsj-ros (commented):
Looks good assuming the CI gives similar results.

PR #63, opened 2021-02-25
updated system_test_cleanup to OpenDDS 3.16+

adamsj-ros:
# Description

This is merging master which includes updates for OpenDDS 3.16+ and some conflict resolution into `system_test_cleanup`.

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] CI tests
- [ ] local tests

# Checklist:

- [x] I have performed a self-review of my own code
- [ ] New and existing system tests pass locally with my changes (when applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -30,7 +30,7 @@ jobs:
     displayName: filesystem check
   - script: |
       cd $(Agent.BuildDirectory)/s
-      git clone https://github.com/oci-labs/rosidl_typesupport_opendds
+      git clone --branch opendds_3.16 https://github.com/oci-labs/rosidl_typesupport_opendds
     displayName: clone rosidl_typesupport_opendds
   - script: |
         cd $(Agent.BuildDirectory)
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -30,7 +30,7 @@ jobs:
     displayName: filesystem check
   - script: |
       cd $(Agent.BuildDirectory)/s
-      git clone --branch opendds_3.16 https://github.com/oci-labs/rosidl_typesupport_opendds
+      git clone --branch opendds_3.16 https://github.com/adamsj-oci/rosidl_typesupport_opendds.git
     displayName: clone rosidl_typesupport_opendds
   - script: |
         cd $(Agent.BuildDirectory)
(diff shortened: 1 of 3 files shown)


rmw_opendds_cpp/src/OpenDDSNode.cpp
```diff
@@ -140,6 +141,9 @@ OpenDDSNode::OpenDDSNode(rmw_context_t & context)
     if (!sub_listener_) {
       throw std::runtime_error("CustomSubscriberListener failed");
     }
+    OpenDDS::DCPS::Discovery_rch disco = TheServiceParticipant->get_discovery(context.options.domain_id);
+    OpenDDS::RTPS::RtpsDiscovery_rch rtps_disco = OpenDDS::DCPS::dynamic_rchandle_cast<OpenDDS::RTPS::RtpsDiscovery>(disco);
+    rtps_disco->use_xtypes(false);
     dp_ = context.impl->dpf_->create_participant(
```
lij-oci, line 147: If the call to rtps_disco->use_xtypes(false) is domain_id specific, not participant specific, these few lines should be moved to rmw_context_impl_t constructor right after TheServiceParticipant->set_default_discovery(OpenDDS::DCPS::Discovery::DEFAULT_RTPS). In that way, it's guaranteed to set xtypes false before creating any participant, and also to avoid calling it repeatedly for each participant.

lij-oci: https://github.com/oci-labs/rmw_opendds/blob/system_test_cleanup/rmw_opendds_cpp/src/init.cpp

after line 24.

lij-oci: That may involve adding domain ID in struct rmw_context_impl_t, since the domain_id is default initialized (RMW_DEFAULT_DOMAIN_ID) in rmw_init_options_t.

lij-oci (approved):

